Issue #13, opened 2016-10-01
indexing tensors


goldingn:
Thanks you so much for this package!

This is a feature request: would it be possible to provide a slimmer syntax for subsetting tensors to match python/tf tensor subsetting, preferably using R's `[`?

i.e., I can do this in Python:
```python
import tensorflow as tf
x = tf.zeros(3)
y = x[1]
sess = tf.Session()
sess.run(y)
```
```
0.0
```
but in R, no dice:
```r
library (tensorflow)
x = tf$zeros(3)
y = x[1]
```
```
Error in x[1] : object of type 'externalptr' is not subsettable
```

I can use other tf functions like `slice` and `gather`, but it's fairly clunky:
```r
y = tf$gather(x, 1L)
sess = tf$Session()
sess$run(y)
```
```
[1] 0
```

jjallaire:
Yes, it should be very straightforward to do this. One thing that's a bit tricky about tensors is that they are 0-based (reflecting the fact that C++ and Python/numpy are 0-based). I'm thinking that the subsetting operators should reflect this as well (otherwise there will be confusion between subsetting functions that are 0-based and operators that are 1-based).

I should be able to get to this early next week. In case you want to play around with this in the meantime the `tf_generics.R` file has a bunch of other customer operators for Tensors: https://github.com/rstudio/tensorflow/blob/master/R/tf_generics.R#L57-L60

goldingn:
Great, thanks! Hopefully I'll l find some time to try implementing this myself. 

I agree, it would be less confusing to keep it 0-indexed. It would be good to flag this to the user, though I'm not sure which doc would be best

jjallaire:
I put a stub in for the generic here: https://github.com/rstudio/tensorflow/commit/3c1c48b1fb877ebd7e6a93d0b3469c0e0203869d

If you could give me a PR for the implementation that would be great.

goldingn:
Some more details of the prototype implementation in the PR:

#### Indexing
Indices are taken to be zero-indexed, inclusive at the lower end and
non-inclusive at the upper end, as in numpy & tensorflow.

#### Syntax
I've attempted to implement a subset of the numpy/tensorflow syntax, with Python's standalone `:` replaced with a blank entry for the default index. I.e. these are equivalent (assuming e.g. a 4x3 Tensor `x`):

python: `x[1, 2]`
R: `x[1, 2]`
(scalar)

python: `x[1:3, 2]`
R: `x[1:3, 2]`
[...]

PR #19, opened 2016-10-03
first pass at subsetting syntax using '['

goldingn:
Here's my first attempt at a solution to #13 
It's not release-ready yet, but there are open design questions (see below) and I'm out of time for today so this PR is just to flag my progress to date. More detailed comments in the issue.

### DIFF
--- a/R/tf_generics.R
+++ b/R/tf_generics.R
@@ -55,10 +55,28 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
   x_size <- x$get_shape()$as_list()
   n_indices <- length(x_size)
 
-  # create a list of the indices provided
-  if (missing(i)) i <- NA
-  if (missing(j)) j <- NA
-  indices <- c(list(i, j), list(...))
+  # capture all indices (skip function, x & drop from arguments)
+  cl <- match.call()
+  args <- as.list(cl)[-(1:2)]
+  indices <- args[names(args) != 'drop']
+
+  # evaluate any calls and replace any skipped indices (names) with NAs
+  indices <- lapply(indices,
+                    function (x) {
+                      if(is.name(x)) NA
+                      else if (is.call(x)) eval(x)
+                      else x
+                    })
+
+  # check all the indices are numeric or NA
+  bad_indices <- sapply(indices,
+                        function (x) !is.numeric(x) & !is.na(x[1]))
+
+  if (any(bad_indices)) {
+    msg <- sprintf('indices %s were not numeric',
+                   paste(which(bad_indices), collapse = ', '))
+    stop (msg)
+  }
 
   n_indices_specified <- length(indices)
 
@@ -75,6 +93,9 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
     missing <- seq_len(n_indices - n_indices_specified) + n_indices_specified
     indices[missing] <- NA
   }
+
+  # strip out any names
+  names(indices) = NULL
 
   # find index starting element on each dimension
   begin <- sapply(indices,
--- a/R/tf_generics.R
+++ b/R/tf_generics.R
@@ -59,6 +59,9 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
     if (!(is.numeric(x) && is.finite(x))) {
       stop ('invalid index - must be numeric and finite')
     }
+    if (!(is.vector(x))) {
+      stop ('only vector indexing of Tensors is currently supported')
+    }
     if (any(x < 0)) {
       stop ('negative indexing of Tensors is not currently supported')
     }
--- a/R/tf_generics.R
+++ b/R/tf_generics.R
@@ -50,9 +50,96 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
 
 #' @export
 "[.tensorflow.python.framework.ops.Tensor" <- function(x, i, j, ..., drop = TRUE) {
-  stop("[ not yet implemented for Tensor")
-}
 
+  # tensor shape as a vector
+  x_size <- x$get_shape()$as_list()
+  n_indices <- length(x_size)
+
+  # create a list of the indices provided
+  if (missing(i)) i <- NA
+  if (missing(j)) j <- NA
+  indices <- c(list(i, j), list(...))
+
+  n_indices_specified <- length(indices)
+
+  # error if too many indices
+  if (n_indices_specified > n_indices) {
+    msg <- sprintf('object has %i dimensions but %i indices were specified',
+                   n_indices,
+                   n_indices_specified)
+    stop (msg)
+  }
+
+  # pad out if too few indices
+  if (n_indices_specified < n_indices) {
+    missing <- seq_len(n_indices - n_indices_specified) + n_indices_specified
+    indices[missing] <- NA
+  }
+
+  # find index starting element on each dimension
+  begin <- sapply(indices,
+                  function (x) {
+                    if (is.na(x[1])) {
+                      0L
+                    } else {
+                      x[1]
+                    }
+                  })
+
+  # find slice size in each dimension (accounting for numpy/tensorflow not
+  # including the last element)
+  slice_end <- sapply(indices,
+                      function (x) {
+                        if (is.na(x[1])) {
+                          Inf
+                        } else {
+                          pmax(x[1], x[length(x)])
+                        }
+                      })
+
+  # crop to Tensor size
+  x_end <- pmax(0, x_size)
+  end <- pmin(x_end, slice_end)
+
+  # convert to shapes
+  begin_shape <- do.call('shape', as.list(begin))
+  end_shape <- do.call('shape', as.list(end))
+
+  # add stride length (always 1) so that the output is consistent with python API
+  stride_shape <- as.list(rep(1L, n_indices))
+
+  # get shrink mask as an integer represent a bytestring
+
+  # if drop=TRUE, drop all *indices* specified as integers,
+  # i.e. for a 2x3 Tensor x:
+  #   x[1:1, ] => shape 1x3
+  #   x[1, ] => shape 3
+  if (drop) {
+    # all
+    shrink_vec <- sapply(indices,
+                         function (x) {
+                           if (!is.na(x[1]) & length(x) == 1) {
+                             1L
+                           } else {
+                             0L
+                           }
+                         })
+  } else {
+    shrink_vec <- rep(0, 4)
+  }
+
+  # convert the (reverse) shrinking bytestring to integer
+  shrink_string <- paste(rev(shrink_vec),
+                         collapse = '')
+  shrink_integer <- strtoi(shrink_string, 2)
+
+  # return the slice
+  tf$strided_slice(input_ = x,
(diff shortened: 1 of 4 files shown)


R/tf_generics.R
```diff
@@ -50,9 +50,96 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+  end_shape <- do.call('shape', as.list(end))
+
+  # add stride length (always 1) so that the output is consistent with python API
+  stride_shape <- as.list(rep(1L, n_indices))
+
+  # get shrink mask as an integer represent a bytestring
+
+  # if drop=TRUE, drop all *indices* specified as integers,
+  # i.e. for a 2x3 Tensor x:
+  #   x[1:1, ] => shape 1x3
```
goldingn: sorry, that should read:
`#   x[1:2, ] => shape 1x3`

R/tf_generics.R
```diff
@@ -50,9 +50,96 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
-}
 
+  # tensor shape as a vector
+  x_size <- x$get_shape()$as_list()
+  n_indices <- length(x_size)
+
+  # create a list of the indices provided
+  if (missing(i)) i <- NA
+  if (missing(j)) j <- NA
+  indices <- c(list(i, j), list(...))
```
jjallaire: We are free to add arguments to an S3 method implementation so I think we could add e.g. `k, l, m, n, o, p` in order to support more empty dimensions.

jjallaire:
Thank you! This looks like a good start.

I think we should also add some brief words about the fact that end indexes are exclusive here: https://github.com/rstudio/tensorflow/blob/master/docs/using_tensorflow_api.Rmd#tensor-indexes

R/tf_generics.R
hadley: We could, but I don't think that's customary (e.g. I don't think the R indexing code for arrays uses more than `i`, `j`, and `k`)

hadley: Since this is general, I would be more tempted to drop the `i`, `j`, and `k`, and just rely on `...`. But the details of how argument passing works for `[` is complicated, so I'd want some example code/unit tests for it.

hadley: Here at least, you should do `indices <- c(i, j, ...)`, and then check that it's a numeric vector

R/tf_generics.R
```diff
@@ -50,9 +50,96 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+  }
+
+  # pad out if too few indices
+  if (n_indices_specified < n_indices) {
+    missing <- seq_len(n_indices - n_indices_specified) + n_indices_specified
+    indices[missing] <- NA
+  }
+
+  # find index starting element on each dimension
+  begin <- sapply(indices,
```
hadley: It would be simpler to use `ifelse`: 

```R
begin <- ifelse(is.na(x), 0L, as.integer(x))
```

R/tf_generics.R
```diff
@@ -50,9 +50,96 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+                      function (x) {
+                        if (is.na(x[1])) {
+                          Inf
+                        } else {
+                          pmax(x[1], x[length(x)])
+                        }
+                      })
+
+  # crop to Tensor size
+  x_end <- pmax(0, x_size)
```
hadley: Is silently cropping the right thing to do? I'd think this should be an error.

R/tf_generics.R
```diff
@@ -50,9 +50,96 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+    shrink_vec <- sapply(indices,
+                         function (x) {
+                           if (!is.na(x[1]) & length(x) == 1) {
+                             1L
+                           } else {
+                             0L
+                           }
+                         })
+  } else {
+    shrink_vec <- rep(0, 4)
```
hadley: Why are these integers and not logicals?

R/tf_generics.R
goldingn: each of `i`, `j`,  `...` can be a vector of length >1, so we don't want to collapse to a numeric vector here.
Will leave it in a list and check each element is a numeric vector.

goldingn: each of `i`, `j`,  `...` can be a vector of length >1, so we don't want to collapse to a numeric vector here.
Will leave it in a list and check each element is a numeric vector.

R/tf_generics.R
goldingn: That's the Python API's (and presumably numpy's) behaviour, so made sense to mirror it here. E.g.:
```python
import tensorflow as tf
x = tf.zeros([2, 2])
x[0:6, :]
```
```
Tensor("strided_slice:0", shape=(2, 2), dtype=float32)
```

R/tf_generics.R
goldingn: The python API uses [`tf.strided_slice`](https://www.tensorflow.org/versions/r0.11/api_docs/python/array_ops.html#strided_slice) for subsetting syntax. Dropping dimensions via `strided_slice` is handled by passing an *integer* whose bitstring represents the mask used to drop dimensions (1 to drop or 0 to keep). E.g. to drop the third element of of a 4d Tensor, the bytestring is `0100` (reverse order) and the integer to pass is `64`, so:
```python
 tf.strided_slice(..., shrink_axis_mask = 64)
```

I'm creating vectors of integers then using `paste` and `strtoi` to create the bitstring and integer in R. I couldn't think of a tidier way of implementing this.

R/tf_generics.R
goldingn: as it's a list and each element has length>1 `ifelse` won't work here.
Something tidier than my `if` code would be very welcome though!

R/tf_generics.R
hadley: Oh yeah, so just do `list(i, j, ...)`

R/tf_generics.R
hadley: Oh hmmmm, you're not going to be able to do much better then, but you should switch from `sapply()` to `vapply()`.  And maybe check the length, rather than `is.na()`?

R/tf_generics.R
hadley: Ah got it

R/tf_generics.R
hadley: A cleverer solution is something like `sum(2^(seq_along(x)-1)[x])`, but that's probably too clever ;)

R/tf_generics.R
goldingn: Ah yeas, that's much cleaner!

R/tf_generics.R
goldingn: yup, that tidies things up  a bit - thanks!

R/tf_generics.R
```diff
@@ -55,10 +55,28 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
   x_size <- x$get_shape()$as_list()
   n_indices <- length(x_size)
 
-  # create a list of the indices provided
-  if (missing(i)) i <- NA
-  if (missing(j)) j <- NA
-  indices <- c(list(i, j), list(...))
+  # capture all indices (skip function, x & drop from arguments)
```
goldingn: This now handles missing indices without needing additional (`k`, `l`, `m` etc.) arguments, and matches R's extract behaviour for arrays. E.g.:

R's array indexing behaviour:
```r
x <- array(NA, dim = c(3, 3, 3, 3, 3, 3))
dim(x[1:2, 1:2, 1:2, 1:2, , 1:2])
```
```
[1] 2 2 2 2 3 2
```
Tensor indexing behaviour as of this commit:
```r
library (tensorflow)
x <- tf$zeros(shape(3, 3, 3, 3, 3, 3))
x[0:2, 0:2, 0:2, 0:2, , 0:2]
```
```
Tensor("StridedSlice:0", shape=(2, 2, 2, 2, 3, 2), dtype=float32)
```
python API indexing behaviour
```python
import tensorflow as tf
x = tf.zeros([3, 3, 3, 3, 3, 3])
x[0:2, 0:2, 0:2, 0:2, :, 0:2]
```
```
Tensor("strided_slice:0", shape=(2, 2, 2, 2, 3, 2), dtype=float32)
```

R/tf_generics.R
```diff
@@ -55,10 +55,28 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+                      if(is.name(x)) NA
+                      else if (is.call(x)) eval(x)
+                      else x
+                    })
+
+  # check all the indices are numeric or NA
+  bad_indices <- sapply(indices,
+                        function (x) !is.numeric(x) & !is.na(x[1]))
+
+  if (any(bad_indices)) {
```
goldingn: Now checking indices are numeric, e.g.:

```r
library (tensorflow)
x <- tf$zeros(shape(3, 3, 3, 3, 3, 3))
x[0:2, 0:2, 0:2, 'apple', , 0:2]
```
```
 Error in `[.tensorflow.python.framework.ops.Tensor`(x, 0:2, 0:2, 0:2,  : 
  indices 4 were not numeric 
```

(message construction could obviously use some work)

goldingn:
Tests coming next.

I still have doubts about the syntax, specifically whether it would be better to have the indices be inclusive at the upper end (but still 0-indexed). Assuming there are no other places in the R API that use the upper end of an index, switching to being inclusive at the upper end would be internally consistent and closer to R syntax. Being R-esque seems preferable to mirroring Python, especially as much of Python's syntax (`begin:`, `:end`, `begin:end:stride`) isn't feasible to implement here anyway.

jjallaire:
I tend to agree that being inclusive at the end index would lead to less
confusion and be closer to the spirit of the R generic. @hadley ?

goldingn:
It would also clear up the inconsistency in behaviour between R's x[1:1, ] (1 row) and Python's x[1:1, ] (0 rows)

R/tf_generics.R
```diff
@@ -50,9 +50,115 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+
+  # tensor shape as a vector
+  x_size <- x$get_shape()$as_list()
+  n_indices <- length(x_size)
+
+  # Capture all indices beyond i and j (skip function, `x`, `drop`, `i` & `j`
+  # from the arguments). This enables users to skip indices to get their defaults
+  cl <- match.call()
+  args <- as.list(cl)[-1]
+  extra_indices <- args[!names(args) %in% c('x', 'i', 'j', 'drop')]
```
hadley: Why can't you just do `list(i, j, ...)` and then check for `NULL`?

jjallaire:
It looks to me like this is nearly ready to go. I added a section on generics to the docs here (https://github.com/rstudio/tensorflow/blob/master/docs/using_tensorflow_api.Rmd#generics). Perhaps we could add the extract operator to the bottom of the table and then just add a few words below about it being zero based, other tips/principles/etc. No more than a paragraph or so + an example or two.

goldingn:
My fork's now up to date, and I've added to the docs.

I didn't mention which features of R's `[` extraction syntax *aren't* implemented (negative, decreasing, block indices) since they seem like edge cases, but the error messages should be fairly informative.

PR #40, opened 2016-11-29
TFLearn DNNClassifier examples

terrytangyuan:
Couple issues noticed:

- hidden_units = c(10L, 15L, 10L) if `L` is not specified, you'll see 
```
Error in py_call(attrib, args, keywords) : 
  ValueError: ('num_outputs should be int or long, got %s.', 10.0)
```
This is caused when `hidden_units` are not of `six.interger_types` in Python, 
```
>>> six.integer_types
(<type 'int'>, <type 'long'>)
```

- My first run of this gives me good accuracy but the runs after give me accuracy like 0.0196 and it's always the same number. This happens even when I re-open my RStudio and even when I change `temp_model_dir`.

### DIFF
--- a/inst/examples/tflearn/README.md
+++ b/inst/examples/tflearn/README.md
@@ -1,0 +1,3 @@
+# Examples using TF.Learn
+
+This directory provides basic examples of TF.Learn to get you started in building deep learning or machine learning models quickly in TensorFlow R. More tutorials in Python can be found [here](https://www.tensorflow.org/versions/master/tutorials/tflearn/index.html#tf-contrib-learn-quickstart) and [here](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/examples/learn). You can mimic those Python examples to write your R program. 
--- a/inst/examples/tflearn/iris_dnn.R
+++ b/inst/examples/tflearn/iris_dnn.R
@@ -18,7 +18,7 @@ classifier <- tf$contrib$learn$DNNClassifier(
   model_dir = temp_model_dir)
 
 # Train a DNN Classifier
-classifier$fit(datasets$data[train_inds, ], datasets$target[train_inds], steps = 200)
+classifier$fit(datasets$data[train_inds, ], datasets$target[train_inds], steps = 100)
 
 # Generate predictiosn on new data
 predictions <- classifier$predict(datasets$data[test_inds, ])
--- a/inst/examples/tflearn/iris_dnn.R
+++ b/inst/examples/tflearn/iris_dnn.R
@@ -1,0 +1,7 @@
+library(tensorflow)
+
+
+datasets <- tf$contrib$learn$datasets$load_dataset("iris")
+feature_columns <- tf$contrib$learn$infer_real_valued_columns_from_input(datasets$data)
+classifier <- tf$contrib$learn$DNNClassifier(feature_columns = feature_columns, hidden_units = c(10, 20, 10), n_classes = 3)
+classifier$fit(datasets$data, datasets$target, steps = 100)
--- a/inst/examples/tflearn/iris_dnn.R
+++ b/inst/examples/tflearn/iris_dnn.R
@@ -1,7 +1,26 @@
 library(tensorflow)
 
+train_inds <- 1:100
+test_inds <- 100:150
+
+temp_model_dir <- tempdir()
 
 datasets <- tf$contrib$learn$datasets$load_dataset("iris")
+
+# Infer the real-valued feature columns
 feature_columns <- tf$contrib$learn$infer_real_valued_columns_from_input(datasets$data)
-classifier <- tf$contrib$learn$DNNClassifier(feature_columns = feature_columns, hidden_units = c(10, 20, 10), n_classes = 3)
-classifier$fit(datasets$data, datasets$target, steps = 100)
+
+# Initialize a DNN classifier with hidden units 10, 15, 10 in each layer
+classifier <- tf$contrib$learn$DNNClassifier(
+  feature_columns = feature_columns,
+  hidden_units = c(10L, 15L, 10L),
+  n_classes = 3L,
+  model_dir = temp_model_dir)
+
+# Train a DNN Classifier
+classifier$fit(datasets$data[train_inds, ], datasets$target[train_inds], steps = 200)
+
+# Generate predictiosn on new data
+predictions <- classifier$predict(datasets$data[test_inds, ])
+accuracy <- sum(predictions == datasets$target[test_inds]) / length(predictions)
+print(paste0("The accuracy is ", accuracy))


jjallaire:
Yes, the need to manually specify `L` is one of the unfortunate side effects of the R to python binding. For shapes we have the `shape` function to smooth things out a bit for a common use case but for the others users do need to keep it in mind. 

The only thing I can think of to do would be to add some sort of API-wide annotations (similar to what we do with help webpages) which would cause us to do the conversion under the hood if necessary.

jjallaire:
We definitely need to get the bottom of the accuracy staying the same! Perhaps there is some sort of subtle floating point comparison issue going on? Maybe run through the script in the debugger to see exactly what values are occurring where.

jjallaire:
I'll merge and we can follow up on the accuracy issue separately.

terrytangyuan:
Yeah I'll keep looking when I get a chance.  Could you try see if you can reproduce it? I suspect it might be caused by [this line](https://github.com/rstudio/tensorflow/blob/dbadad2c626c197c85d4f6b47bdc7bb7aa9abff5/src/python.cpp#L802) so the classifier overfitted or maybe I wasn't very awake last night so the first accuracy I got never existed.

ShantanuKumar:
The accuracy thing is because of the fact that the data is ordered, and you are picking training set with only 1s and 0s, whereas the test set has all 2s. Random sampling helps.

terrytangyuan:
That's very good catch and I wasn't vey awake indeed! Thank you, will try later!

terrytangyuan:
Wanna submit a fix? Perhaps better with stratified sampling.

ShantanuKumar:
The other issue I found was if you change the model in anyway, like number of hidden units or layers, it will throw error. With changing model, the `model_dir` should be changed also. `tempdir` is per session in R. So running different models in the same session won't work. Error message is not helpful at all in that regard.
Sorry, I can't submit a fix at the moment. If you want, please do it.

jjallaire:
Yes, to address the tempdir() issue just change that call to tempfile()
then do a dir.create on the result.

terrytangyuan:
Yep I am aware of that. I manually changed dirs when I realized the second issue. Could you clarify the first issue you encountered? I remember I changed hidden units last night and I didn't have any issues.

ShantanuKumar:
I get this error
`Error in py_call(attrib, args, keywords) : basic_string::_M_replace_aux` when I try to change the number of hidden units in any layer, after I have built a model, in the same R session. If I change the `model_dir`, it works. So if you change hidden units, after building a model, and don't change the `model_dir`, you will get this error.

terrytangyuan:
Yes that's the right behavior since if you provide model for that's existed it will load the existing one but your model structure is different. I am not sure why the error message is so obscure though. @jjallaire I also noticed this issue where the error messages are truncated or stack trace is not detailed. It would be better to have full stack trace for easier debugging purposes.

terrytangyuan:
@ShantanuKumar BTW that's one of the current ways of persisting a TFLearn model.

ShantanuKumar:
ok. I have just started using TensorFlow.

I also get this warning message
`Instructions for updating:
The default behavior of predict() is changing. The default value for
as_iterable will change to True, and then the flag will be removed
altogether. The behavior of this flag is described below.
WARNING:tensorflow:Calling predict (from tensorflow.contrib.learn.python.learn.estimators.estimator) with as_iterable=False is deprecated and will be removed after 2016-09-15.`
Then, when I set `as_iterable=TRUE` in `predict`, warning goes away, but it returns a generator object which is useless in R.

terrytangyuan:
@jjallaire Is there any special treatment for iterators?

jjallaire:
Not currently, but I'm sure we could get this to work. We either need to
just return the raw generator (assuming there is a sensible way to interact
with it from R) or we need to convert it to an R list. If you guys could
provide me with the simplest possible TF/python expression that yields a
generator I'll take a closer look.

jjallaire:
It looks like Python generators automatically implement the iterator
protocol (https://docs.python.org/3/c-api/iter.html). This means that if we
support iterable we'll get support for generator. We won't need a special
construct if it's okay to just convert iterable to list (i.e. drain the
source and place all the items in an R list). Is that acceptable behavior
in this case or do we need the incremental yielding afforded by generators?

jjallaire:
Try the following with the return value:

```r
py_builtin <- import("__builtin__")
result <- <your call to TFLearn>
result <- py_builtin$list(result)
```

If that gives you what you need then we can make that automatic whenever we encounter an iterable or a generator.

@bwlewis In case he has come up with another better way to handle this.

terrytangyuan:
Yeah that's fine as one option for now but Iterables are mostly needed for
streaming data. I think if we only support returning the resulted list it
defeats the purpose of that so we may want to allow two ways for users to
handle this.

jjallaire:
Okay, then I think we should also add an R function for streaming processing of iterators. Something like the `iapply` function described here: https://cartesianfaith.com/2013/01/05/infinite-generators-in-r/

terrytangyuan:
Yeah this approach sounds good to me. We should leave this option to users
instead of implicitly converting them.

jjallaire:
I've implemented support for converting to R lists/vectors from Python
iterators/generators here:
https://github.com/rstudio/tensorflow/commit/33062417a80c8c0375ea8922c9ffffd052902c3c

Now I'll take a look at a streaming version

On Sun, Dec 4, 2016 at 11:46 AM, Yuan (Terry) Tang <notifications@github.com

jjallaire:
Here's an issue to track work on streaming for generators: https://github.com/rstudio/tensorflow/issues/47

Note that the non-streaming version should now be working with the last commit.

jjallaire:
Note that we currently do an automatic conversion to a list or vector from the iterator. Once we have the streaming version we may want to not convert and create generics for as.list, as.vector to make the copying conversion explicit.

terrytangyuan:
Awesome! That's super quick!

jjallaire:
One question: is the dominant use case here streaming? (i.e. is the whole reason that an iterator/generator is being returned rather than a list that its better to fetch it incrementally?). In that case it's definitely better to force explicit conversion to an R list or vector, however if you *usually* just want a copied list or vector we may want to make iteration require an explicit opt-in. Let me know what you think....

jjallaire:
It seems like maybe generators should be automatically streamed and iterators should perhaps be automatically copied?

terrytangyuan:
Good points. I agree with what you said above. Perhaps we should just let
users specify as options() to switch this.

On Mon, Dec 5, 2016 at 10:01 AM JJ Allaire <notifications@github.com> wrote:

It seems like maybe generators should be automatically streamed and
iterators should perhaps be automatically copied?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rstudio/tensorflow/pull/40#issuecomment-264875972>, or mute
the thread

jjallaire:
I've changed the approach. There is now a single `iterate` function which can be used to apply an R function to each item within an iterator/generator: https://rstudio.github.io/tensorflow/using_tensorflow_api.html#iterators

PR #41, opened 2016-11-29
Fix temp_model_dir, added notes for sampling, added to test

terrytangyuan:
Quick/lazy fix and added to test


terrytangyuan:
Thanks for the tip on `tempfile()` ;-) @jjallaire

Issue #44, opened 2016-12-02
Input function doesn't work in R


ShantanuKumar:
https://www.tensorflow.org/versions/r0.11/tutorials/input_fn/index.html
I have been following the tutorial for using input function, but I can't seem to make it work. 
Since we don't have multiple return values in R, I am using list. Has anyone been able to implement the code as mentioned in the tutorial? 

Thank you for your help!

terrytangyuan:
Could you paste your code here?

terrytangyuan:
This should be fixed in master.

PR #45, opened 2016-12-05
Custom decay example and case handling for closure object

terrytangyuan:
@jjallaire This doesn't work yet since closure object conversion is not implemented. Could you help work on the implementation so this example could run? 

Related issue: https://github.com/rstudio/tensorflow/issues/44

### DIFF
--- a/inst/examples/tflearn/iris_custom_decay_dnn.R
+++ b/inst/examples/tflearn/iris_custom_decay_dnn.R
@@ -1,0 +1,40 @@
+library(tensorflow)
+
+optimizer_exp_decay <- function() {
+  global_step <- tf$contrib$framework$get_or_create_global_step()
+  learning_rate <- tf$train$exponential_decay(
+    learning_rate = 0.1,
+    global_step = global_step,
+    decay_steps = 100,
+    decay_rate = 0.001)
+  return(tf$train$AdagradOptimizer(learning_rate=learning_rate)) # print should contain more info
+}
+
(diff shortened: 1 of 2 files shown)


terrytangyuan:
It is for this case but sometimes we need to pass a function that returns multiple items. For example:

```
def my_input_fn()

    # Preprocess your data here...

    # ...then return 1) a mapping of feature columns to Tensors with
    # the corresponding feature data, and 2) a Tensor containing labels
    return feature_cols, labels
```

This function could return multiple items and this `my_input_fn()` will be passed to `fit()`, like this:

```
classifier.fit(input_fn=my_input_fn, steps=2000)
```

jjallaire:
That should still work fine. I'm assuming that `my_input_fun` will be written in R and will return an R list with 2 elements. The python function we synthesize to wrap it will call the R function then convert the 2-element R list to a Python tuple. Does that make sense?

terrytangyuan:
Yes that makes sense. Right now there are only cases with 1-3 elements being returned.

jjallaire:
I'm going to merge this PR then continue working on getting support for closures working on master.

terrytangyuan:
Thanks!

jjallaire:
I just checked in initial support for closures -- your example now seems to work!

Still need to do some polishing and write some tests but the basics should be there. Give it a try with some less trivial examples and let me know what you find.

terrytangyuan:
Great! That was quick!

PR #50, opened 2016-12-06
Fix edge case for iterator/generator conversion

terrytangyuan:
Fixed the following edge case:

This works
```
test <- import("tftools.test")
> as.integer(iterate(test$makeGenerator(5)))
[1] 0 1 2 3 4
```
But this doesn't
```
test <- import("tftools.test")
> a <- test$makeGenerator(5)
> a
<generator object makeGenerator at 0x119ae6050>
Python iterator/generator (use tensorflow::iterate to traverse)
> b <- iterate(a)
> b
list()
```

### DIFF
--- a/R/python.R
+++ b/R/python.R
@@ -234,7 +234,8 @@ with.tensorflow.builtin.object <- function(data, expr, as = NULL, ...) {
 iterate <- function(x, f = base::identity) {
   if (!inherits(x, "tensorflow.builtin.iterator"))
     stop("iterate function called with non-iterator argument")
-  invisible(py_iterate(x, f))
+  invisible(result <- py_iterate(x, f))
+  return(result)
 }
 
 #' @export
--- a/tests/testthat/test-python-iterators.R
+++ b/tests/testthat/test-python-iterators.R
@@ -11,4 +11,9 @@ test_that("Iterators reflect values back", {
 
 test_that("Generators reflect values back", {
   expect_equal(as.integer(iterate(test$makeGenerator(5))) + 1L, seq(5))
+  a <- test$makeGenerator(5)
+  b <- as.integer(iterate(a))
+  expect_equal(b + 1L, seq(5))
 })
+
+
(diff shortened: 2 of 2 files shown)


jjallaire:
I think this is by design (generators are "drained" when they are called). Try adding the following to the bottom of the test.py file and you'll see that the python code does the same thing as the R code:

```python
g = makeGenerator(5)

for i in g:
  print i

for i in g:
  print i
```

terrytangyuan:
Oops I neglected the nature of Python iterator. I was focusing from users perspective - when users have something like 
```
predictions <- classifier.predict()
```
and they realize this is a iterator and then runs `iterate(predictions)` after seeing the suggestion from `print(predictions)`. They will be confused when they see `list()` as the result.

jjallaire:
Okay, let me see if there's a way for me to make that less confusing. Will post back here when I have a solution.

jjallaire:
Okay, no I'm unable to repro the problem you originally pointed out. Here's the code I'm using:

```r
library(tensorflow)
test <- import("tftools.test")
a <- test$makeGenerator(5)
a

b <- iterate(a)
b
```

That still prints `b` for me. Are you still seeing the problem?

terrytangyuan:
I see. I accidentally did this instead:

```
> library(tensorflow)
> test <- import("tftools.test")
> a <- test$makeGenerator(5)
> a
<generator object makeGenerator at 0x1160aa7d0>
Python iterator/generator (use tensorflow::iterate to traverse)
> 
> b <- iterate(a)
> b <- iterate(a)
> b
list()
```

I iterated over `a` twice. I think this aligns with the design (make sure old pointer is drained). This could probably be a good test to include as well. 

I am closing this now. Apologize for the confusion.

jjallaire:
Test added here: https://github.com/rstudio/tensorflow/commit/a1bed771f0a38c48e8c6efdede574a6002300b15

terrytangyuan:
👍

PR #61, opened 2017-01-03
Typo

jimhester:
My previous PR had a typo which caused the builds to fail, this should fix it, but I would wait until the builds finish successfully before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ addons:
       - python-joblib
       - python-dev
 
-script: -|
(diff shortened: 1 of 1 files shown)


jjallaire:
This one failed as well. I'll merge anyway as we can't hardly get *worse* than we already are. I'll also add you as a committer so you can experiment directly on master.

kevinushey:
I think the issue here is spelling (tens*e*rflow vs. tens*o*rflow)

PR #67, opened 2017-01-11
automatically simplify results of iterate to vector when appropriate

jjallaire:
@terrytangyuan I saw that you had an `unlist` call wrapping the results of `iterate` and thought we could automagically eliminate the need to do that in many cases. Take a look and let me know if you think this will present any problems with the way iterators are used in TF.


terrytangyuan:
Sounds good to me

PR #70, opened 2017-01-20
initial stab at 'linear_regression' wrapper for tensorflow

kevinushey:
This PR ports most of the ideas used in `sparklyr` for implementing a `tensorflow` wrapper to linear regression. The goal is for the user to be able to write something like:

    fit <- tf_linear_regression(mpg ~ cyl, data = mtcars)
    fit <- mtcars %>% tf_linear_regression(mpg ~ cyl)

to fit a linear regression model. Note that methods for e.g. `summary()`, `predict()` and so on are not yet implemented in this PR.

Is this the approach we want to take for some easy-to-use wrappers to TensorFlow modeling? If so, do we want to tweak this interface? If not, do we just want slightly simpler building blocks?

FWIW, I feel like there's a couple kinds of top-level APIs we could expose in this package:

1. The 'mimic an R interface' approach, as done in `sparklyr` and explored in this PR,
2. The 'use pipes to build layers', similar to the Keras approach,
3. Something intermediate between 1. and 2.


R/preconditions.R
```diff
@@ -0,0 +1,148 @@
...
+#' @name ensure
+#' @export
+ensure_scalar_character <- make_ensure_scalar_impl(
+  is.character,
+  "a length-one character vector",
+  as.character
+)
+
+
+require_file_exists <- function(path, fmt = NULL) {
```
terrytangyuan: Are the functions below needed?

R/preconditions.R
```diff
@@ -0,0 +1,148 @@
...
+    )
+  }
+
+  object
+}
+
+#' Enforce Specific Structure for R Objects
+#'
+#' These routines are useful when preparing to pass objects to
+#' a Spark routine, as it is often necessary to ensure certain
```
terrytangyuan: Change this to TensorFlow?

R/tf_linear_regression.R
```diff
@@ -0,0 +1,80 @@
...
+
+  # Extract response vector and convert to one-dimensional matrix
+  rv <- as.matrix(x[response], ncol = 1L)
+
+  # Extract features
+  k <- length(features)
+  fv <- as.matrix(x[features], ncol = k)
+
+  # Construct tensorflow placeholders
+  X <- tf$placeholder("float", shape = shape(NULL, k))
```
terrytangyuan: Seems okay to use low level API as a start but later we will probably need to switch to use TF.Learn for this

internal/tf-linear-regression.R
```diff
@@ -0,0 +1,10 @@
+library(tensorflow)
+
```
terrytangyuan: Is this file needed?

kevinushey: Sorry; this was supposed to be `.gitignore`d!

R/preconditions.R
kevinushey: They're not (were accidentally brought in while porting these utilities from `sparklyr`). I'll clean them up!

R/preconditions.R
kevinushey: Oops; indeed!

R/tf_linear_regression.R
kevinushey: I'll see if I can make that happen before merging this PR.

terrytangyuan: No worries. I'll also need test a few things out to figure out how to best design for flexibility vs. usability

R/tf_linear_regression.R
```diff
@@ -0,0 +1,81 @@
...
+#'
+#' @template roxlate-tf-x
+#' @template roxlate-tf-response
+#' @template roxlate-tf-features
+#' @template roxlate-tf-intercept
+#' @template roxlate-tf-dots
+#' @template roxlate-tf-options
+#'
+#' @export
+tf_linear_regression <- function(x,
```
jjallaire: I'm wondering how we should choose names for these functions. It's tempting to just use `linear_regression` as the code will be nicer to read. That said, I don't know if the grammar of modeling will end up wanting to use that function? It's also possible based on our conversation the other day that the grammar of modeling will use `tf_linear_regression` :-\

We could adopt the TFLearn nomenclature and call it `LinearRegressor`, `tf_linear_regressor`, or `linear_regressor`

@hadley @topepo

terrytangyuan: My two cents: 

1. with tf to avoid signature collisions
2. adopt TF.learn nomenclature since if we use regression, we cannot really have estimation, right?

hadley: I think it's likely we'll end up with a `linear_regression` function (or more likely `mod_linear()`) that would call `tf_linear_regression()` so the UI pain will go away in the long term.

jjallaire:
So tf_linear_regressor ?

On Fri, Jan 20, 2017 at 5:52 PM, Yuan (Terry) Tang <notifications@github.com

jjallaire:
A thought: I think that as long as we are actively iterating on the nature of our tflearn interfaces we should separate them into a separate package. This is because I think it will take a while for us to finally agree on an interface that captures both high and lower-level interaction, has the right factoring of code visa-vi keras, and is synched w/ Max and Hadley's work. In the meantime I'd like to submit the lower-level tensorflow package to CRAN soon and I don't want to commit to a public interface for tflearn that we might end up needing to evolve. 

I propose we create a `tflearn` package which sits on top of the `tensorflow` package. We can merge it in later if and when we are confident that we've got the final interface nailed.

topepo:
I've been looking into some of the pre-processing methods too to (eventually) include as steps for recipes but would probably make them available in a different package so avoid excessive dependencies.

terrytangyuan:
I agree that this package should stay low level with all the flexibilities
it provides. We can build more friendly interfaces on top of this.Though I
am not sure tflearn is a good name given the unfortunate naming collision
with the other unrelated tflearn/tflearn package.

terrytangyuan:
I think it's also good timing to submit to CRAN given the first major
version of TF is being released soon.

kevinushey:
How about just `tensorlearn` instead of `tflearn` for the package name?

terrytangyuan:
That's not bad. I also had toy package before called rflow that can be reused if you think it's good. Or learnflow...trying to be creative here...

jjallaire:
I think tflearn is a fine name since the overlap already exists on the
python side and given the state/standing of both projects it's extremely
unlikely that another R wrapper will be written for the other tflearn.

On Sat, Jan 21, 2017 at 3:46 PM, Yuan (Terry) Tang <notifications@github.com

terrytangyuan:
sounds good to me

kevinushey:
Now iterating on the [`tflearn`](https://github.com/rstudio/tflearn) repository.

PR #76, opened 2017-01-27
Dynamic Loading for Python Bindings

jjallaire:
This PR changes the approach to binding to the Python C API from installation/compilation time to runtime. This means that:

1. The compiled version of the tensorflow package can bind to multiple versions of python (including both v2 and v3).

2. We can distribute compiled binaries from CRAN (previously users would have needed to always install the package from source).

We've also implement various heuristics for scanning the system for a version of python where tensorflow is installed (e.g. checking `/usr/local/bin/python` and other conventional installation paths, inspecting the windows registry, etc.). This should make it much more straightforward to get the package working even in the presence of multiple versions of python or the "wrong" version of python being on the PATH as is often the case on OS X.

Finally, the `tensorflow:::tf_config` function exposes detailed information about the version of python discovered, versions of numpy and tensorflow therein, etc. If the package fails to discover tensorflow at load time then appropriate error messages and the tf_config are printed as package startup messages when the package is attached.


terrytangyuan:
Great work! This is super helpful!

terrytangyuan:
Should we add test env for Windows too?

jjallaire:
Are there Windows test environments on travis? If so then definitely yes.

terrytangyuan:
Never mind. I don't think so.

eddelbuettel:
Folks who can bear the pain (ie: not me) use Appveyor for this.  Kirill and others have a lot of examples.

PR #81, opened 2017-01-30
Fix #80 (find PyMalloc libpython)

### LINKED ISSUE #80: Package fails to find PyMalloc libpython
Gentoo builds Python to use pymalloc, and as per [PEP 3149](https://www.python.org/dev/peps/pep-3149/), the Python 3 libpython shared libraries, are suffixed with `m`, e.g. `/usr/lib/libpython-3.5m.so`. However, it seems the package only looks for `libpython3.5.so` without any possible suffix.

Indeed,
```R
library(tensorflow)

Python shared library '/usr/lib64/libpython3.5.so' not found, Python bindings not loaded.

<snip>

Detected Python configuration:

python:         /usr/bin/python-exec2c
libpython:      /usr/lib64/libpython3.5.so[NOT FOUND]
version:        3.5.2 (default, Oct  2 2016, 22:41:13)  [GCC 5.4.0]
numpy:          /home/alex/.local/lib64/python3.5/site-packages/numpy
numpy_version:  1.11.3
tf:             /home/alex/.local/lib64/python3.5/site-packages/tensorflow
```

NullInfinity:
Search for `libpython` now finds libraries of the form `libpython3.5.so`
and `libpython3.5m.so` including the `m` suffix for PyMalloc Python as
per PEP 3149.

In the event that libraries are found with and without the suffix, the
one without is used, so there should be no change for existing
RTensorFlow users with working setups.

Note: [PEP3149](https://www.python.org/dev/peps/pep-3149/) mentions other suffices besides `m`, which I have not addressed (e.g. `u` for wide Unicode support). These could be added too if desired, but I omitted them for now for simplicity.

### DIFF
--- a/R/tf_config.R
+++ b/R/tf_config.R
@@ -87,10 +87,14 @@ tf_python_config <- function(python, python_versions) {
     python_libdir_config <- function(var) {
       python_libdir <- config[[var]]
       ext <- switch(Sys.info()[["sysname"]], Darwin = ".dylib", Windows = ".dll", ".so")
-      libpython <- file.path(python_libdir, paste0("libpython" , version, ext))
+      libpython <- file.path(python_libdir, paste0("libpython" , version, c("", "m"), ext))
+      libpython <- libpython[file.exists(libpython)]
+      if (length(libpython) == 0)
+        return(NA)
+      libpython[1]
     }
     libpython <- python_libdir_config("LIBPL")
-    if (!file.exists(libpython))
+    if (is.na(libpython))
       libpython <- python_libdir_config("LIBDIR")
   }
 


NullInfinity:
Oops, this doesn't build just yet. Will re-open when fixed.

jjallaire:
What are the values of these configuration variables on you system:

```
print(sysconfig.get_config_var('EXT_SUFFIX'))
print(sysconfig.get_config_var('SOABI'))
```

They are `None` on my system. It sounds to me like we might be able to tease out whether the `m` is necessary based on the presence of these variables.

NullInfinity:
```python
>>> import sysconfig
>>> print(sysconfig.get_config_var('EXT_SUFFIX'))
.cpython-35m-x86_64-linux-gnu.so
>>> print(sysconfig.get_config_var('SOABI'))
cpython-35m-x86_64-linux-gnu
```

PR #100, opened 2017-03-02
implement a simple argument parser

kevinushey:
This PR implements a simple argument parser, intended to be compatible with the way that Python's `argparse` parses arguments. The intention is for this to be used in cases where application receive command line arguments that might otherwise be parsed by Python's `argparse`. See e.g.: https://github.com/GoogleCloudPlatform/cloudml-samples/blob/34959ec732ac8efa7fb3efdd293ac7956fa5253a/mnist/hptuning/trainer/model.py#L44-L51

Since many sample TensorFlow applications seem to use `argparse` for command line arguments, I think it makes sense to provide a similar helper in this package -- let me know if you think this belongs elsewhere.


jjallaire:
It does seem like we will need to provide some facility for this since most examples do something like this. I wonder though if given that we know we have Python >= 2.7 underneath whether we should just implement this on top of `argparse` ?

PR #102, opened 2017-03-09
Update tutorial_mnist_beginners.Rmd

danklotz:
When I tried `tf$global_variables_initializer()`did not exist. Could it be that `tf$initialize_all_variables()` is the needed function?


jjallaire:
It depends on the version of tensorflow you are using (yours is probably not recent enough). Here are the docs on the deprecation: https://www.tensorflow.org/api_docs/python/tf/initialize_all_variables

PR #116, opened 2017-05-17
Accepts "Linux" as a supported name in is_ubuntu

cbarbu:
This is necessary as recent versions of ubuntu use "Linux in uname.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -14,7 +14,7 @@ is_osx <- function() {
 
 is_ubuntu <- function() {
   uname <- suppressWarnings(system2("uname", "--all", stdout = TRUE, stderr = TRUE))
-  grepl("ubuntu", uname[[1]], ignore.case = TRUE)
+  grepl("ubuntu|Linux", uname[[1]], ignore.case = TRUE)
 }
 
 


jjallaire:
I think we might need a more fine grained check than that. TensorFlow binaries are officially only supported on Ubuntu. I'm worried that looking for "Linux" will allow installation to proceed on other Linux platforms.

Is there a more fine grained check available that will work on all versions of Ubuntu?

jjallaire:
Hopefully this will do it: https://github.com/rstudio/tensorflow/commit/1be61bb647f649dc0ea653db0bfe6b5df3aa79c6

terrytangyuan:
The answer in [this SO question](https://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name) seems good.

jjallaire:
I think this code should do it:

https://github.com/rstudio/tensorflow/blob/master/R/utils.R#L15-L24

I could call `lsb_release` directly, but that will just end up reading from the same /etc/lsb-release file we are accessing. Let me know if you think there's any reason we should be explicitly calling lsb_release for just Ubuntu detection (for multi-distro detection I agree it would be better/required).

terrytangyuan:
That sounds good to me!

PR #132, opened 2017-06-11
Enhancements to `tensorboard()` and updated DESCIPTION

terrytangyuan:
Added `reload_interval` and `purge_orphaned_data` that will be useful in many scenarios.


jjallaire:
To build the docs in this project you should use:

```r
source("tools/docs/build.R")
```

As that does some additional stuff required by our docs.

jjallaire:
I'm actually wrong. That's keras that needs that. Looks good as is!

jjallaire:
Actually, this project doesn't use pkgdown for docs at all, rather we maintain a website here: https://github.com/rstudio/tensorflow.rstudio.com

So I'd just remove the `docs` directory from this PR.

terrytangyuan:
Gottcha. Just removed that folder!

terrytangyuan:
I don't have access to that repo. Do I have to update the docs there?

jjallaire:
We should probably add a section to the article in the other repo (just gave you access).

PR #140, opened 2017-06-26
handle undefined dimensions in extraction; fixes #139

### LINKED ISSUE #139: unexpected behavior when extracting from tensors with undefined shape
Suppose I have a placeholder like `x` define below and I want to extract the first column:

```
x <- tf$placeholder(tf$int16, shape = list(NULL, 1L))
x[,0L]
```
This returns the following error:

```
Error in pmin(end, x_size) : 
  (list) object cannot be coerced to type 'double'
```

The equivalent python code works fine:

```
x = tf.placeholder(tf.int16, [None, 1])
x[:,0]
<tf.Tensor 'strided_slice:0' shape=(?,) dtype=int16>
```
Also the following works as expected:

```
> x <- tf$placeholder(tf$int16, shape = list(1L, 1L))
> x[,0L]
Tensor("StridedSlice:0", shape=(1,), dtype=int16)
```

goldingn:
implements a bugfix and adds a test

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -46,6 +46,8 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
   # check for blank spaces in the call
   is.blank <- function (x) is.name(x) && as.character(x) == ''
 
+  integer_mask <- function (mask) sum(2 ^ (seq_along(mask) - 1)[mask])
+
   # evaluate any calls (in the environment calling `[`) and replace any
   # skipped indices (blank names) with NAs
   evaluate_index <- function (x) {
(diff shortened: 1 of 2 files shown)


R/generics.R
```diff
@@ -76,6 +78,13 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
 
   # tensor shape as a vector
   x_size <- x$get_shape()$as_list()
+
+  # if it has dimension(s) of undefined size, this will be a list with NULLs
+  if (is.list(x_size)) {
+    x_size <- lapply(x_size, function (x) ifelse(is.null(x), NA, x))
```
terrytangyuan: Do you mean NULL or NA here?

goldingn:
If `x` is a placeholder with known shape, `x$get_shape()$as_list()` returns an integer vector.

If `x` has one or more unknown dimensions `x$get_shape()$as_list()` returns a list with `NULL` for those dimensions of unknown size. The rest of the code is expecting a vector, so this line converts `NULL`s to `NA`s so that x_size can be converted into a vector.

terrytangyuan:
I see now. Thanks for the clarification!

PR #142, opened 2017-06-30
Upgrade pip before pip installing tensorflow

fritzo:
Old versions of pip fail to locate the tensorflow python package, and old versions of setuptools fail to handle python wheels. This is problematic e.g. in old travis-ci environments with ancient apt versions of pip. To fix this, we `pip install --upgrade pip` and `pip install --upgrade setuptools` before `pip install tensorflow`.

Test status:
- Passes: locally on Linux
- Passed: on travis-ci Linux https://travis-ci.org/nimble-dev/nimble/jobs/248861441
- Passed: on travis-ci OS X https://travis-ci.org/nimble-dev/nimble/jobs/248861446

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -302,10 +302,9 @@ install_tensorflow_virtualenv <- function(python, virtualenv, version, gpu, pack
   } else {
     cat("Using existing virtualenv at ", virtualenv_path, "\n")
   }
+  virtualenv_bin <- function(bin) path.expand(file.path(virtualenv_path, "bin", bin))
 
   # upgrade pip so it can find tensorflow
-  virtualenv_bin <- function(bin) path.expand(file.path(virtualenv_path, "bin", bin))
-  pkgs <- tf_pkgs(version, gpu, package_url)
   cmd <- sprintf("%ssource %s && %s install --ignore-installed --upgrade %s%s",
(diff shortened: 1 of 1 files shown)


fritzo:
FYI We are adding a tensorflow back-end to the NIMBLE compiler, and we very much appreciate the installation and integration work y'all have done!

jjallaire:
It looks to me as if these installs are done into the virtualenv so don't ever require sudo, is that correct?

Are these installation commands basically free after the first time they run? (i.e. they do version checks and no-op if the update is already there).

jjallaire:
Here's another version of this which factors the pip invocation into a common helper function: https://github.com/rstudio/tensorflow/pull/143

I'll close this PR in lieu of the new one. Thanks so much for this contribution! I was struggling to get travis working with TF and now I know why! (and can re-enable travis :-))

fritzo:
Thanks, JJ!

Re: travis, I'm still having trouble getting R to find the tensorflow package, but I'll let you know if I find a fix.

>  https://travis-ci.org/nimble-dev/nimble/jobs/248984951#L1883
>  Python module tensorflow was not found.

jjallaire:
It took me a while to debug this on travis, but here is a working
.travis.yml for the tensorflow package:

https://github.com/rstudio/tensorflow/blob/master/.travis.yml

PR #145, opened 2017-07-01
Add `tf_imperative()` for imperative style TensorFlow

terrytangyuan:
Allows users to easily experiment with TensorFlow using imperative style without needing to set up sessions and graphs.


jjallaire:
The `with` construction creates some nice scoping but also makes it hard to use from the REPL (will only work inside scripts). I noticed that the Python docs emphasize a different usage pattern: https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/imperative

Which in R would be:

```r
tf <- tf$contrib$imperative
```

I wonder if as long as this library is in contrib we might be better off documenting this usage pattern rather than adding a feature to the package?  Another possibility would be to create a separate `tfimperative` package that we could then merge into the core tensorflow package if/when this gets promoted out of contrib?

Feature suggestion for `with_imperative`:  what would you think of converting tensor return values of the `expr` to R by evaluating the tensor? That would allow easy assignment of the results to an object that can then be used conveniently from R. 

This is incredibly useful. If you think it's nearly certain that this will get promoted out of contrib then I'd still be okay adding it to the tensorflow package. Your call.

terrytangyuan:
I actually have implemented the feature suggestion too but felt like it's not a good fit here (I was experimenting the idea I suggested through Slack)! 

I am okay to move this to a separate package then and keep this core package pure and always backwards compatible but it's not too much code yet. Maybe we can mark this as experimental or generate a warning when being called?

jjallaire:
Why didn't you like how the feature suggestion turned out? My concern is that when you call this you end up with a tensor object that's not terribly useful on the R end.

terrytangyuan:
I think you are right. I'll probably that in soon. I was thinking whether users will always want that. Maybe I can add a eval = TRUE argument?

jjallaire:
How about convert = TRUE?

terrytangyuan:
Sounds good to me.

jjallaire:
I'm okay including this in the tensorflow package as long as the docs say
it's experimental (we don't need a warning).

terrytangyuan:
Great, thanks!

terrytangyuan:
@jjallaire Just pushed some changes suggested. Could you take a look?

R/imperative.R
```diff
@@ -0,0 +1,60 @@
...
+  convert = TRUE,
+  env = parent.env(environment())
+) {
+  server <- tf$python$training$training$Server$create_local_server()
+  target <- server$target
+  with(tf$contrib$imperative$imperative_mode$ImperativeMode(target), as = imperative_mode, {
+    if (new_step) {
+      with(imperative_mode$new_step(), as = new_step, {
+        invisible(expr)
+        if (convert) invisible(implicit_eval_tensor(env = env))
```
jjallaire: I was think that the way this would work would be to just convert the result of the expression to R, e.g.

```r
res <- force(expr)
if (inherits(res, "tensorflow.python.framework.ops.Tensor")
   invisible(res$eval())
else
   invisible(res)
```

I think that converting all of the objects could get way too expensive and might be somewhat surprising.

R/imperative.R
```diff
@@ -0,0 +1,60 @@
+#' Context manager for imperative style TensorFlow
+#' 
+#' The results of the computation are available right after the execution of a line of code and
+#' can be converted to base R objects automatically.
+#' 
+#' Note that this function is currently only experimental, meaning that the interface is subject
+#' to change or remove at later releases.
+#'
```
jjallaire: Should we also document running an entire interactive session in imperative mode, e.g.

```r
tf <- tf$contrib$imperative

# Do a bunch of work with tf

# Call tensor$eval() if you want to get the results back to R at the end
```

R/imperative.R
terrytangyuan: Good point. Maybe I am overthinking here but what if users want multiple results from the scope (maybe for performance optimization)? If we do this, they'll have to use with_imperative() multiple times? I am thinking maybe an argument named "convert_intermediate_result" to eval all the tensors or "convert_names" to convert a list of variables with the names.

R/imperative.R
terrytangyuan: Yes I'll also document this.

R/imperative.R
jjallaire: They can always call tensor$eval() to capture an R object result back into the parent scope, e.g. at the end of the block:

```r
x <- x$eval()
```

I think that if we document that the return value is automatically converted and that other objects can be converted as illustrated above that we should be covered.

terrytangyuan: Sounds good. Will do.

R/imperative.R
```diff
@@ -0,0 +1,76 @@
+#' Context manager for imperative style TensorFlow
```
jjallaire: Maybe a different title now that we've rename the function?

R/imperative.R
```diff
@@ -0,0 +1,76 @@
...
+
+convert_tensors <- function(res, convert) {
+  convert_tensor <- function(res, convert) {
+    if (convert && inherits(res, "tensorflow.python.framework.ops.Tensor"))
+      invisible(res$eval())
+    else
+      invisible(res)
+  }
+  if (is.list(res)) {
+    lapply(res, function(item) {
```
jjallaire: Should we preserve names if it's a named list?

terrytangyuan: I think it's preserved. Check the last test?

R/imperative.R
terrytangyuan: Updated!

terrytangyuan:
Updated!

R/imperative.R
jjallaire: Sure enough it is, thanks!

PR #148, opened 2017-07-08
Add functions for dimensionality retrieval for tensors

terrytangyuan:


R/generics.R
```diff
@@ -37,6 +37,26 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
...
+"dim.tensorflow.python.framework.ops.Tensor" <- function(x) {
+  x$get_shape()$as_list()
+}
+
+#' @export
+"length.tensorflow.python.framework.ops.Tensor" <- function(x) {
+  Reduce(`*`, dim(x))
+}
+
+#' @export
```
jjallaire: nrow and ncol and actually not generics so we can't do this (however, the implementation of `nrow` and `ncol` just calls `dim` so will still work on tensors without this code.

jjallaire: i.e. you can just delete this code (it's not being called anyway)

terrytangyuan:
Thanks. That's good to know! Just removed them.

PR #152, opened 2017-07-12
provide 'parse_flags()' for parsing + unifying YAML configuration + command line arguments

kevinushey:
This PR adds a `config()` function, which:

1. Reads in a configuration from a `config.yml` file,
2. Parses and merges in any command-line arguments, and
3. Applies a user-specified R hook (if any) to the resulting configuration.

cc: @jjallaire; in particular I'm curious what you think about the user hook approach here. (e.g. recall that in `cloudml`, we added a bit of extra tooling to make sure that `source()`-based workflows would work correctly, and that required some dynamic patching of the `job_dir`).

Closes #152.


R/config.R
```diff
@@ -0,0 +1,37 @@
...
+#' @example examples/examples-config.R
+config <- function(config = Sys.getenv("R_CONFIG_ACTIVE", unset = "default"),
+                   file = "config.yml",
+                   arguments = commandArgs(TRUE))
+{
+  # read configuration file
+  config <- config::get(config = config, file = file)
+
+  # merge parsed command line arguments
+  config <- config::merge(config, tensorflow::parse_arguments(arguments))
```
jjallaire: I think R CMD check will complain about the explicit qualification of `parse_arguments`

examples/examples-config.R
```diff
@@ -0,0 +1,27 @@
+if (!requireNamespace("yaml", quietly = TRUE))
```
jjallaire: Maybe this code should actually be the basis of a test rather than an example?

R/config.R
```diff
@@ -0,0 +1,37 @@
...
+
+  # apply a post-config hook
+  hook <- getOption("tensorflow.config.hook")
```
jjallaire: I'd say I'm leaning against having any hook mechanism at all, as it would be great if to write scripts for cloudml training you didn't need any special setup (e.g. `library(cloudml)`). I'm thinking we should go back to something like this:

```yaml
default:
   job_dir: !expr unique_dir("jobs")
```

or if you don't care about a unique directory for each job:

```yaml
default:
   job_dir: "job"
```

Then of course when we are running on CloudML the `--job-dir` command line option would override this.

My core goal here is to be able to establish a standard local training workflow that can then be deployed to CloudML (or other distributed training environments) with no code changes. To me requiring that a hook be installed means that scripts will need something like `library(cloudml)` which will defeat the objective of having no special code modifications required.

kevinushey: I'm trying to recall the motivation we had around this before -- in `cloudml`, we actually decided _not_ to include `job_dir` within the configuration file, and instead had the notion of a `project_name` and `job_output` configuration fields. See:

https://github.com/rstudio/cloudml/blob/master/examples/census/config.yml

Then, jobs were constructed with unique 'names' generated based on the application name + timestamp, and then the job directory was constructed by concatenating the `job_output` and `job_name` fields.

https://github.com/rstudio/cloudml/blob/master/R/config.R#L29-L34

So at least there, we didn't want to specify `job_dir` in the configuration file; instead, we wanted it generated dynamically based on other configuration values. To accomplish that, we had a 'hook' that would populate it when necessary (as it was in `source()`-based workflows).

In other words, the hook is useful when you want to dynamically populate configuration fields based on other configuration fields. We might be able to get away with leaving it out for now but we may want to re-visit it later.

kevinushey:
I've removed the bit about the hook for now, but I do think we'll need to revisit it.

Because deployments will require some kind of entrypoint, we can always try to populate the configuration hook in there (or transform configuration as needed for a particular deployment target). Recall that we did this sort of thing in `cloudml`:

https://github.com/rstudio/cloudml/blob/master/inst/cloudml/cloudml/deploy.R

Note that this doesn't require any intervention by the user; it just provides some mechanism for us to communicate with the application configuration during deployments.

examples/examples-config.R
kevinushey: I think this is okay as an example -- it's a test that is also useful for users to see.

jjallaire: It just seems like if we want to show an illustrative example then maybe a directory with an actual config file and a script that reads it would be better than writing a dummy config to a temp file (I don't think a user would ever do that)

kevinushey: Okay, I'll make that change.

jjallaire:
See the Google Doc I just sent around: I think we can handle this via the
injection of a $RUNDIR environment variable in the same spot where we used
to populate the hook.

jjallaire:
LGTM! Let me know when it's ready to merge

kevinushey:
Should be good to go!

PR #153, opened 2017-07-13
Check the actual Python path during tensorflow_install() 

missaugustina:
The previous method assumes python exists in /usr/bin or
/usr/local/bin which would cause the python_unix_binary()
method to fail if the python to be used lives anywhere else (eg,
local install in the home directory). This change uses the "which"
command to determine the location of the python binary.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -357,12 +357,10 @@ install_tensorflow_windows_system <- function(python, pip, version, gpu, package
 
 
 python_unix_binary <- function(bin) {
-  locations <- file.path(c("/usr/bin", "/usr/local/bin"), bin)
-  locations <- locations[file.exists(locations)]
-  if (length(locations) > 0)
-    locations[[1]]
-  else
-    NULL
+  tryCatch({
+      system2("which", bin, stdout=TRUE, stderr=TRUE)
+    }, warning = function(x){NULL}
+  )
 }
 
 python_version <- function(python) {


missaugustina:
I was hoping to trigger Travis to test this but it looks like you don't have the hook set up?

missaugustina:
Just using Travis to test this :)

jjallaire:
A couple of issues here:

1. You can just use the R `Sys.which` function to interrogate the PATH.

2. We've actually been down this road before (e.g. see this commit from a couple of weeks ago which was reverted: https://github.com/rstudio/tensorflow/commit/892e23ca1d62655664f73a3443e8e03d4b048909). The current approach where we don't inspect the PATH is an attempt to be as conservative as possible in locating Python (i.e. strongly prefer the system version of Python, which has the most testing with tensorflow installations). We actually ran into problems interrogating the PATH with finding other versions of Python (e.g. homebrew on the Mac) which in turn had configuration incompatibilities with tensorflow.

Definitely open to making a change here if warranted, but we have a few months of more or less trouble-free installation under our belts with this approach so I don't want to change things lightly.

jjallaire:
To further clarify, in the PR you say "Tools like pyenv or local installs of Python may fail". That's true, but in this case we are specifically attempting to sidestep e.g. an existing pyenv installation, because we are going to go ahead and create a new virtuaenv anyway. In other words, we aren't really looking for the Python environment to install tensorflow into, but rather a widely tested with tensorflow Python binary that we can bootstrap a virtualenv with

missaugustina:
Closing per comments above. Opened Issue #154 to address these concerns and prevent others from going down this path!

PR #166, opened 2017-08-19
poll for availability of tensorboard using http rather than stdout

jjallaire:
Tensorboard shipped with TensorBoard v1.3 doesn't write a newline at the end of it's output and as a result p$read_output_lines() hangs. This works around that issue by using http to poll for availability of tensorboard.

Tested on Mac, Windows, and Ubuntu with TF v1.2 and 1.3

@kevinushey Could you review this before we merge it?

### DIFF
--- a/R/tensorboard.R
+++ b/R/tensorboard.R
@@ -167,37 +167,37 @@ launch_tensorboard <- function(log_dir, host, port, explicit_port, reload_interv
                                "--purge_orphaned_data", purge_orphaned_data),
                              stdout = "|", stderr = "|")
 
-  # poll for output until we've succesfully started up or the process dies
+  # poll for availability of the http server (continue as long as the
+  # process is still alive). note that we used to poll for stdout however
+  # tensorflow v1.3 stopped writing a newline after printing the host:port
+  # and caused us to haning in p$read_output_lines()
   started <- FALSE
+  Sys.sleep(0.25)
+  conn <- url(paste0("http://", host, ":", as.character(port)))
   while(!started && p$is_alive()) {
+    Sys.sleep(0.25)
+    tryCatch({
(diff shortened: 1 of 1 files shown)


R/tensorboard.R
```diff
@@ -167,37 +167,37 @@ launch_tensorboard <- function(log_dir, host, port, explicit_port, reload_interv
...
+  # and caused us to haning in p$read_output_lines()
   started <- FALSE
+  Sys.sleep(0.25)
+  conn <- url(paste0("http://", host, ":", as.character(port)))
   while(!started && p$is_alive()) {
+    Sys.sleep(0.25)
+    tryCatch({
+      readLines(conn, n = -1)
+      started = TRUE
+      close(conn)
```
kevinushey: Maybe just use `on.exit(close(conn), add = TRUE)` above after creating the connection?

R/tensorboard.R
```diff
@@ -167,37 +167,37 @@ launch_tensorboard <- function(log_dir, host, port, explicit_port, reload_interv
...
+    },
+    error = function(e) {},
+    warning = function(w) {}
```
kevinushey: FWIW the use of a warning handler here implies that `started = TRUE` will never be hit if `readLines()` returns a warning. E.g. as in

```R
tryCatch({
  warning("warning")
  print("hello")
}, warning = function(w) print("caught warning")
)
```

The line `print("hello")` is never executed. Are those the intended semantics?

kevinushey:
Mostly just minor comments (I think it's okay to ignore them if you prefer) but otherwise LGTM!

jjallaire:
Made both changes. Let me know if all looks well and I'll merge.

kevinushey:
LGTM!

PR #174, opened 2017-09-03
Random seed setter and session creator for reproducibility

terrytangyuan:
https://github.com/rstudio/keras/issues/42


jjallaire:
Thanks! I think I am going to start with a method in the Keras package (which can also handle Theano and perhaps CNTK if there is a way to do it there). Then once we confirm that works well I'll migrate functions back to reticulate / tensorflow over time. I want to do it this way so as not to create a set of cascading dependencies for the keras dev version right before we announce the package.

terrytangyuan:
Sounds good. Should I close this one for now?

jjallaire:
Yes let's close this one for now.

jjallaire:
Related PRs in Keras are here: https://github.com/rstudio/keras/pulls

PR #185, opened 2017-10-09
Move export_savedmodel into tf from tfestimators

javierluraschi:
Moves `export_savedmodel` into `tensorflow`, see https://github.com/rstudio/tfestimators/pull/95.


jjallaire:
I'm going to merge this since `export_savedmodel` is the clear default candidate for the function name. We can discuss realtime later on this week whether to rename to `save_model` or something else.

jjallaire:
You'll want to bump the version of the tensorflow package in this PR, and then depend on that version (and add rstudio/tensorflow to the Remotes: in the DESCRIPTION). I'll merge and then bump the version.

javierluraschi:
Sounds good!

PR #186, opened 2017-10-10
Add support for view_savedmodel()

javierluraschi:
Adds `view_savedmodel()` as a companion function to `export_savedmodel()` since once the models are exported, it is usually desired to be able to visualize them to compare and understand them further.

Usage:

```r
export_path <- export_savedmodel(model)
view_savedmodel(export_path)
```

For instance, I used this while troubleshooting `keras` vs `tensorflow` model exports:

graph-large_attrs_key _too_large_attrs limit_attr_size 1024 run logs 2f 1


R/save.R
```diff
@@ -13,3 +13,48 @@
...
+#' @param export_dir The path to the exported directory, as a string.
+#' @param log_dir Directories to scan for training logs. If this is a named
+#'   character vector then the specified names will be used as aliases within
+#'   TensorBoard.
+#'
+#' @return The path to the exported directory, as a string.
+#'
+#' @export
+view_savedmodel <- function(
+  export_dir = file.path(getwd(), "trained/tensorflow-mnist/1/saved_model.pb"),
```
terrytangyuan: Remove the hard-coded path?

R/save.R
```diff
@@ -13,3 +13,48 @@
...
+
+  with(tf$Session() %as% sess, {
+    with(gfile$FastGFile(export_pb, "rb") %as% f, {
+      graph_def <- tf$GraphDef()
+
+      data <- compat$as_bytes(f$read())
+      sm <- saved_model_pb2$SavedModel()
+      sm$ParseFromString(data)
+
+      if (sm$meta_graphs$`__len__`() > 1) stop("Saved model contains more than one graph")
```
terrytangyuan: Should we try fix this in reticulate so that we can use `length()`?

jjallaire: I just tried to do this and discovered that the RStudio IDE uses the `length()` function on all objects in the global environment. This unfortunately results in errors for Python objects which don't have a `__len__` method. I think the best we can do is implement `length()` for objects which we know have the method (e.g the implementation for tuple here: https://github.com/rstudio/reticulate/blob/master/R/python.R#L553-L558)

R/save.R
jjallaire: Yes, this should just be a required parameter without a default.

R/save.R
```diff
@@ -13,3 +13,48 @@
...
+#' @param log_dir Directories to scan for training logs. If this is a named
+#'   character vector then the specified names will be used as aliases within
+#'   TensorBoard.
+#'
+#' @return The path to the exported directory, as a string.
+#'
+#' @export
+view_savedmodel <- function(
+  export_dir = file.path(getwd(), "trained/tensorflow-mnist/1/saved_model.pb"),
+  log_dir = file.path(getwd(), "logs")
```
jjallaire: It seems like this should just always be a temp directory rather than "logs" (I don't think we want artifacts within the current working directory just because we viewed a model)

R/tensorboard.R
```diff
@@ -34,7 +34,8 @@
 #'   terminate TensorBoard.
 #'
 #' @export
-tensorboard <- function(log_dir, action = c("start", "stop"),
+tensorboard <- function(log_dir = file.path(getwd(), "logs"),
```
jjallaire: The default `log_dir` needs to remain `NULL` so that we can do automatic resolution for tf_runs. In practice, using a "logs" dir within the current working directory isn't really a thing except for in demo code, as you can't really log events over and over again into the same directory (i.e. if you really wanted to use a directory named "logs" your script would have to repeatedly delete it).

R/save.R
```diff
@@ -13,3 +13,48 @@
...
+
+      g_in <- tf$import_graph_def(sm$meta_graphs[[0]]$graph_def)
+    })
+
+    train_writer <- tf$summary$FileWriter(log_dir)
+    train_writer$add_graph(sess$graph)
+    train_writer$close()
+  })
+
+  if (interactive() && !tensorboard_is_running()) tensorboard(log_dir = log_dir)
```
jjallaire: I think the call to `interactive()` isn't necessary (since the user has explicitly requested the view action we can execute it without qualification). 

Also, you don't need the tensorboard_is_running check here as the design of the tensorboard function is that it will kill any existing tensorboard session when a new one is requested.

R/save.R
javierluraschi: Thanks, fixed.

R/save.R
javierluraschi: Opened https://github.com/rstudio/reticulate/issues/111

R/save.R
javierluraschi: Makes sense, changed.

R/tensorboard.R
javierluraschi: Reverted.

R/save.R
javierluraschi: Changed.

R/save.R
```diff
@@ -13,3 +13,48 @@
...
+#' @param log_dir Directories to scan for training logs. If this is a named
+#'   character vector then the specified names will be used as aliases within
+#'   TensorBoard.
+#'
+#' @return The path to the exported directory, as a string.
+#'
+#' @export
+view_savedmodel <- function(
+  export_dir,
+  log_dir = tempdir()
```
jjallaire: It seems to me like the `log_dir` argument should be gone entirely and always be the result of `tempfile()` (as opposed to `tempdir()` which gives the root of the R session temp directory). I can't think of a reason a user would want to control this and/or keep it around after the view is done.

R/save.R
```diff
@@ -13,3 +13,48 @@
...
+#' @param export_dir The path to the exported directory, as a string.
+#' @param log_dir Directories to scan for training logs. If this is a named
+#'   character vector then the specified names will be used as aliases within
+#'   TensorBoard.
+#'
+#' @return The path to the exported directory, as a string.
+#'
+#' @export
+view_savedmodel <- function(
+  export_dir,
```
jjallaire: Maybe name this `saved_model_dir` or `model_dir`?

R/save.R
javierluraschi: K!

PR #187, opened 2017-10-13
Support installing nightly versions on PyPI directly

terrytangyuan:
`pip install tf-nightly` and `pip install tf-nightly-gpu` are available on PyPI now.


jjallaire:
This is exactly how I would have done it. I'll push some changes to show you a different approach.

jjallaire:
@terrytangyuan The main change I made was to rename `package_url` to `packages` and then to automatically supply both the "tf-nightly" and "tensorflow-tensorboard" packages when the version is "nightly"

terrytangyuan:
Looks much better! Thanks!

Issue #111, opened 2017-04-21
Unexpected behavior when extracting from tensors with higher dimension


dfalbel:
I'm creating a function subset_tensor define with:

```
subset_tensor <- function(x, a, b){
  x[,0:(a-2), 0:(b-2),]
}
```

Then I apply it to a placeholder tensor:

```
k <- tf$placeholder(tf$float32, shape = c(1L, 600L, 600L, 3L))
subset_tensor(k, 600, 600)
#> Error in eval(expr, envir, enclos) : object 'b' not found
```

Note that it works if call directly:

```
k[,0:(600-2), 0:(600-2),]
Tensor("StridedSlice_5:0", shape=(1, 599, 599, 3), dtype=float32)
```

When I define a function `subset_tensor2` for a 2 dimension tensor, It works fine.

```
subset_tensor2 <- function(x, a, b){
  x[0:(a-2), 0:(b-2)]
}
subset_tensor2(l, 600, 600)
#> Tensor("StridedSlice_6:0", shape=(599, 599), dtype=float32)
```

Am I doing something wrong?

jjallaire:
This must have to do with the way evaluation is performed. @goldingn Do you have any insight as to why the function call based version wouldn't work here?

goldingn:
Good catch @dfalbel!

Yes, the evaluation of extract calls is a bit of a nightmare. The generic extract function has arguments: `i, j, ... , drop = TRUE` - i.e. the first two indices named and subsequent ones aren't. In order to interpret missing indices as meaning 'all entries' in the higher dimensions, the calls for those indices are caught and then evaluated later.

Turns out the bug you found is just because those calls are evaluated in the wrong environment, hence 'b' could not be found. I've just submitted [a PR](https://github.com/rstudio/tensorflow/pull/112) to fix this, and added a simple test. Using the version in that PR, your example works fine for me:

```r
devtools::install_github('goldingn/tensorflow@bb32c5d')

subset_tensor <- function(x, a, b){
  x[,0:(a-2), 0:(b-2),]
}

k <- tf$placeholder(tf$float32, shape = c(1L, 600L, 600L, 3L))
subset_tensor(k, 600, 600)
#> Tensor("StridedSlice_5:0", shape=c(1, 599, 599, 3), dtype=float32)
```

[...]

PR #193, opened 2017-11-01
support additional dimensions in extract syntax

jjallaire:
@goldingn I noticed that I was unable to use variables in dimensions that weren't explicitly supported by the extract syntax (e.g. the following would give an error: `x[,,,index]` whereas the following would work fine: `x[,,,0]`

This PR adds support for 2 additional dimensions. I tried to follow the pattern in the code you originally submitted but it's possible that I got something wrong. It's also possible there is a better way to enable using variables for index dimensions. Let me know what you think.

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -62,7 +62,7 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
 
 
 #' @export
-"[.tensorflow.python.framework.ops.Tensor" <- function(x, i, j, ..., drop = TRUE) {
+"[.tensorflow.python.framework.ops.Tensor" <- function(x, i, j, k, l, ..., drop = TRUE) {
 
   # check for blank spaces in the call
   is.blank <- function (x) is.name(x) && as.character(x) == ''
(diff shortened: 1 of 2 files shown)


goldingn:
Thanks @jjallaire - the original code is intended to capture variable indices, even in dimensions beyond those named. And in fact it does work if they are more complicated expressions (as in #111), it just seems to be some simple error based on how expressions are detected.

So this doesn't work (even with this patch, since it's on the 5th dimension):
```r
x <- tf$constant(array(dim = rep(3, 5)))
index <- 2
x[, , , , index]
```
```
Error in validate_index(x) : invalid index - must be numeric and finite 
```
but this does (!):
```r
x[, , , , (index)]
```
```
Tensor("StridedSlice_4:0", shape=(3, 3, 3, 3), dtype=bool)
```
so the bugfix should be even simpler than this. I'll take a look now.

goldingn:
Oh, that didn't take long. [This](https://github.com/rstudio/tensorflow/blob/feature/more-extract-dims/R/generics.R#L77) just needed to change from:
```r
    else if (is.call(x))
```
to
```r
    else if (is.call(x) | is.name(x))
```

With this fix there is no need for extra indices, and since they deviate from the generic arguments. It's probably worth rolling back to just `i` and `j`, unless there's some other motivation for extra indices.

I have a fork of this branch [here](https://github.com/goldingn/tensorflow/tree/feature/more-extract-dims) where I have reverted the commit above, fixed the bug and added a test. You could merge that into this PR before merging into master, to record this discussion?

jjallaire:
Thanks!! Merged and merged. Much simpler indeed :-)

Issue #195, opened 2017-11-02
Consider 1-based indexing for tensor [ operator


jjallaire:
I've recently been working on some Keras examples that mix use of R/NumPy arrays and raw TF tensor objects. Within the code the same data is accessed via the `[` operator in both contexts. However, one context is 0-based and the other is 1-based. 

This makes for an incredibly confusing situation, and that's for me the maintainer of all of these packages! I wonder if we should revisit our decision to use 0-based indexing for the  tensor`[` operator. The original motivation was two-fold:

1) The other lower-level `tf$` functions use 0-based indexing; and

2) If you were porting Python code that used 0-based indexing then this would ease the porting.

However, I wonder if the case of mixed R array and TF tensor access will be more common. I think it's a reasonably clean story to say that *as a convenience* we make TF tensors 1-based when using the R `[` operator. This is in fact more semantically consistent with `[`, in which AFAICT 0 is equivalent to dropping a dimension.

If we did this we could detect 0 based indexes and yield an error from them (with a note describing how to put the package in backward compatibility mode via an option). I realize that this would cause some short-term pain, but if 1-based is really the right long term direction it's much better that we do it now.

@goldingn I am particularly interested in your opinion about this.

jjallaire:
Here's the original discussion for reference: https://github.com/rstudio/tensorflow/pull/19

Note that we are already inconsistent with Python by being end-inclusive. It seems like being fully consistent with R `[` would be more clear. No matter what we do some people will be confused some of the time, but to me I can't think of anything as bad as writing code that mixes 0 and 1 based indexing!

@dfalbel Would be interesting to get your take here as well.

goldingn:
Yeah, I suspect being fully consistent with R will increase uptake among R users. I don't think we'll be able to detect and warn on all 0-based indices, but even if it's a breaking change, I think it'll be worth it in the long run.

There are a lot of weird quirks to R's extract and replace syntax (e.g. using logicals; extracting by cell number; extracting by a matrix with `dim` columns), and people seem to use it in very different ways. So replicating all of that manually would be a ton of work.

[...]

PR #199, opened 2017-11-05
R-like extraction for tensors

goldingn:
This implements a (slightly cleaned up) port of greta's extraction behaviour as a solution to #198.

It should behave *exactly* like extraction on an R array, such that if a user creates an array `x` and extracts elements to an array `y` like this:
```r
y <- x[<some extraction call>]
```
then this should have an identical result:
```r
x_tf <- tf$constant(x)
y_tf <- x_tf[<some extraction call>]
y <- tf$Session()$run(y_tf)
```

There are tests to this effect, for all the extraction syntax that springs to mind. 

This retains 0-based indexing as an option, as in #195, but under a different name to make clear it uses the new default is complete R-like extraction syntax, not just 1-based indexing.

### DIFF
--- a/R/extraction.R
+++ b/R/extraction.R
@@ -1,0 +1,200 @@
+# both options for tensor extraction, mimicing R's behaviour, or using 0-based
+# indexing like python, but without python's R-incompatible syntax (e.g.
+# strides)
+
+# tensor extraction mimicing R
+extract_r_like <- function(x, call) {
+
+  dims_in <- dim(x)
+
+  # create a dummy array containing the order of elements Python-style
+  dummy_in <- dummy(dims_in)
+
+  # modify the call and environment to swap dummy_in for the target, and use
+  # primitive subsetting
+  call_list <- as.list(call)[-1]
+  env <- as.list(parent.frame(n = 2))
+  env[[call_list[[1]]]] <- dummy_in
+  dummy_out <- with(env, do.call(.Primitive("["), call_list))
+
+  # coerce result to an array
+  dummy_out <- as.array(dummy_out)
+
+  # get number of elements in input and dimension of output
+  nelem <- prod(dims_in)
+  dims_out <- dim(dummy_out)
+  dims_out <- do.call(shape, as.list(dims_out))
+
+  # get the index in flat python format, as a tensor
+  index <- flatten_rowwise(dummy_out)
+
+  # flatten tensor, gather using index, reshape to output dimension
+  tensor_in_flat <- tf$reshape(x, shape(nelem))
+  tf_index <- tf$constant(as.integer(index), dtype = tf$int32)
+  tensor_out_flat <- tf$gather(tensor_in_flat, tf_index)
+  tensor_out <- tf$reshape(tensor_out_flat, dims_out)
+
+  tensor_out
+
+}
+
+# convert an array to a vector row-wise
+flatten_rowwise <- function (array) {
+  dim <- dim(array)
+  array <- aperm(array, rev(seq_along(dim)))
+  dim(array) <- NULL
+  array
+}
+
+# convert an vector to an array row-wise
+unflatten_rowwise <- function (array, dim) {
+  array <- as.array(array)
+  dim(array) <- rev(dim)
+  array <- aperm(array, rev(seq_along(dim)))
+  dim(array) <- dim
+  array
+}
+
+# create an array with the same dimensions as tensor and fill it with
+# consecutive increasing integers in python order
+dummy <- function (dims) {
+  vec <- seq_len(prod(dims)) - 1
+  unflatten_rowwise(vec, dims)
+}
+
+# ~~~~~~
+
+# tensor extract syntax with zero-based indexing
+extract_zero_based <- function (x, i, j, ..., drop = TRUE) {
+
+  # tensor shape as a vector
+  x_size <- x$get_shape()$as_list()
+
+  # if it has dimension(s) of undefined size, this will be a list with NULLs
+  if (is.list(x_size)) {
+    x_size <- lapply(x_size, function (x) ifelse(is.null(x), NA, x))
+    x_size <- unlist(x_size)
+  }
+
+  n_indices <- length(x_size)
+
+  # Capture all indices beyond i and j (skip function, `x`, `drop`, `i` & `j`
+  # from the arguments). This enables users to skip indices to get their defaults
+  cl <- match.call()
+  args <- as.list(cl)[-1]
+  extra_indices <- args[!names(args) %in% c('x', 'i', 'j', 'drop')]
+
+  # if i wasn't specified, make it NA (keep all values)
+  if (missing(i)) i <- list(NA)
+  else i <- list(validate_index(i))
+
+  # if j wasn't specified, but is required, keep all elements
+  # if it isn't required, skip it
+  if (missing(j)) {
+    if (n_indices > 1) j <- list(NA)
+    else j <- list()
+  } else {
+    j <- list(validate_index(j))
+  }
+
+  # evaluate and fill in blanks
+  extra_indices <- lapply(extra_indices, evaluate_index)
+
+  # combine the indices & strip out any names
+  indices <- c(i, j, extra_indices)
+  names(indices) <- NULL
+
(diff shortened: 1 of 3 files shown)


goldingn:
I haven't attempted to spot 0-based indices here, because 0s are allowed in indices for R's array extraction syntax (those elements are just ignored):

```r
x <- array(1:3)
x[0:2]
```
```
[1] 1 2
```
and in the new TF implementation
```r
x_tf <- tf$constant(x)
y_tf <- x_tf[0:2]
tf$Session()$run(y_tf)
```
```
[1] 1 2
```

jjallaire:
This looks awesome, definitely the way to go! Very much appreciated.

I realize your time is tight and I'm the deadline motivated one so I am happy to make any further tweaks discussed here after the PR is merged to master.

I realize that the 0-based syntax is valid in R but it seems mostly unnecessary (since it's ignored) so I'm wondering if emitting a warning for some period of time (e.g. a year) might be valuable? We could have this warning NOT fire if the user explicitly sets `options(tensorflow.r_like_extract = TRUE)` as opposed to receiving the default value.

goldingn:
Yeah, a warning like that sounds sensible!

jjallaire:
Is there a way to do that in your greta-implementation? (it looks like the indexes might be somewhat opaque). If there is any way whatsoever to do it I think we should as it will save users lots of time and trouble). I am happy to implement this if you should give me a pointer as to how.

goldingn:
I imagine the best bet would be to crib code from the 0-indexing function - that grabs indices for each dimension which could be checked for 0s.

That code is super clunky and opaque because of having to parse missing arguments in the dots, but for a first pass you could just copy-paste all the code up until the strided_slice call, then lool for 0s in those indices. We can always tidy it up later.

jjallaire:
@goldingn I'm noticing that some Keras tensor indexing code that used to work correctly is now failing with the revised implementation:

```r
library(keras)

model <- keras_model_sequential() %>% 
  layer_dense(units = 64, input_shape = c(784)) %>% 
  layer_dense(units = 10, activation = 'softmax')

model$output[,1]
```

Gives:

```
Error in prod(dims) : invalid 'type' (list) of argument
```

Whereas this used to give:

```
Tensor("StridedSlice:0", shape=(?,), dtype=float32)
```

So there is a failure mode we may not have thought about here. If this is too intracted perhaps we should consider modifying the old code to pass the tests and then migrating the greta code only when we have more time. Let me know what you think.

goldingn:
I won't be able to get to a laptop for a bit to check this out, but my immediate guess would be that whatever class `model$output` is, its `dim()` method is behaving differently to simple tensors.

That could be cleared up either in keras, or by doing `dims_in <- unlist(dim(x))` at the top of `extract_r_like()`

jjallaire:
The class is no different that a regular TF tensor:

```r
> class(model$output)
[1] "tensorflow.python.framework.ops.Tensor"      "tensorflow.python.framework.ops._TensorLike"
[3] "python.builtin.object"    
```

```r
> class(tf$constant(42))
[1] "tensorflow.python.framework.ops.Tensor"      "tensorflow.python.framework.ops._TensorLike"
[3] "python.builtin.object"    
```

goldingn:
Huh, OK thanks for checking that out!
I'll take a look in the morning.

jjallaire:
Okay, thanks! 

I've further narrowed the issue to the handling of placeholder tensors. I can also reproduce the error with:

```r
library(tensorflow)
x <- tf$placeholder(tf$float32, shape = shape(NULL, 100))
x[,1]
```
`Error in prod(dims) : invalid 'type' (list) of argument`

I think this is likely to the fact that the placeholder dimension is `NULL`:

```
> dim(x)
[[1]]
NULL

[[2]]
[1] 100
```

Which results in an error when this expression is evaluated:

```
> prod(list(NULL, 100))
Error in prod(list(NULL, 100)) : invalid 'type' (list) of argument
```

jjallaire:
The simple error with the `tf$placeholder` is fixed by this commit: https://github.com/rstudio/tensorflow/pull/200/commits/2dfe92910a66720cafc6bcbb42a9d498ba9b2070

Which basically just substitutes 1 for any dimensions that are `NULL`

Unfortunately the Keras model code now breaks with:

```
 Error in `[[<-`(`*tmp*`, call_list[[1]], value = dummy_in) : 
  invalid subscript type 'language' 
```

jjallaire:
A great clue, this code actually works:

```r
output <- model$output
output[,1]
```

So a simplier repro that fails would be:

```r
library(tensorflow)
container <- list(x = tf$constant(matrix(c(1:4), ncol = 2, nrow = 2)))
container$x[,1]
```

So the greta style indexing isn't handling list elements. Hopefully there is a simple fix for this!

jjallaire:
In terms of cribbing code from the 0 indexing function, it looks like I would just need to crib code until `begin` is defined rather than all the way up the strided slice call, is that correct?

goldingn:
I had forgotten about placeholders with unknown dimensions. That's incompatible with the strategy for full R-like extraction, but was already implemented in the 0-based extraction.

Note that setting unknown dimensions to 1 clears that particular error, but returns the wrong result. It should have shape `(?, )` or `(?, 1)` with `drop=TRUE`.

In #202 I've fixed this by using 'manual' 1-based extraction if there are unknown dimensions, and full R-like extraction otherwise.

I've also fixed the invalid subscript type error, and had a ho at detecting 0s in the indices

dfalbel:
@jjallaire @goldingn 
It seems that this added a lot of overhead in extraction. See this benchmark for example:

```
> options(tensorflow.r_like_extract = FALSE)
> library(tensorflow)
> 
> x <- tf$placeholder(dtype = tf$float16, shape = c(3, 25, 71, 512))
> 
> microbenchmark::microbenchmark(
+   x[1,,,], times = 10
+ )
Unit: milliseconds
       expr      min       lq     mean   median       uq     max neval
 x[1, , , ] 2.496407 2.542692 5.116753 2.626734 3.194224 26.3578    10

Restarting R session...

> library(tensorflow)
> 
> x <- tf$placeholder(dtype = tf$float16, shape = c(3, 25, 71, 512))
> 
> microbenchmark::microbenchmark(
+   x[1,,,], times = 10
+ )
Unit: milliseconds
       expr      min       lq     mean   median       uq      max neval
 x[1, , , ] 894.8528 918.0925 940.6862 924.0265 936.2776 1025.532    10
```

jjallaire:
Wow, that is a huge difference indeed.

I wonder if should consider implementing 1-based indexing for the `tf$strided_slice` implementation after all? I realize it's a bit more work for us but it seems like it could save users quite a bit of CPU time!

Here's the implementation in progress for that: https://github.com/rstudio/tensorflow/pull/198

It has 3 tests failing and I'm not quite sure how to resolve them. If we can get those tests to pass I could move the warning code over from the existing implementation.

@goldingn Interested in your take on whether this performance difference is actually worth living with for the additional syntactic flexibility. At initial glance it seems to me like 1 second is too long for a straightforward slicing operation.

goldingn:
Yeesh, that is slow. 
I can do some profiling to see what the time suck is, though I suspect it's all the `tf$reshape()`ing. 
I'm actually more concerned about any slow down when executing the graph, rather than constructing it - we should definitely benchmark that.

Assuming the slow down is the `tf$reshape()`ing, the easiest fix is to drop full R-like extracting from the API, and use the 'manual' `tf$strided_slice()` one-based indexing that's already implemented for tensors with unknown dimensions. It should work fine when all dimensions are known too.

The `[` method in `generics.R` uses conditionals to select between the three extraction methods (zero-based manual, one-based manual, or R like). So it would be easy to change that and delete the unnecessary tests.

I'd be happy to do this, but won't have time until next week at the earliest, so very happy for someone else to implement instead if this is urgent!

goldingn:
I think speed is probably more important than flexibility here - people using tensorflow will know that they'll need to use a slightly different API than for arrays.

That's slightly different for greta, where identical behaviour to arrays is important. Over there I'll try to detect all-numeric, positive, increasing indices, and use the manual extraction in that case.

jjallaire:
PR here: https://github.com/rstudio/tensorflow/pull/203

PR #202, opened 2017-11-06
Feature/r like extract

goldingn:
* Fix container subsetting issue
* handle placeholders with unknown dimensions
* try to spot 0-indexing and warn when appropriate


jjallaire:
Thanks, wonderful!  

Apologize for the fire drill here but there's nothing like going into print to motivate making sure that APIs are exactly right for the long terrm :-)

PR #203, opened 2017-11-09
don't use r-like extraction from greta

jjallaire:
@dfalbel pointed out here (https://github.com/rstudio/tensorflow/pull/199#issuecomment-342887212) that the feature complete version of r-like extraction doesn't perform nearly as well as the previous implementation. This PR uses the existing implementation (enhanced to optionally have a 1 rather than 0 basis) rather than the greta implementation.

@goldingn Let me know if this handles things correctly / as you imagined.

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -68,25 +68,12 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
   check_zero_based(call)
 
   r_like_extract <- getOption("tensorflow.r_like_extract", TRUE)
-  unknown_dim <- unknown_dimension(x)
 
-  if (r_like_extract && !unknown_dim) {
-
-    # if we can, extract exactly as in R
-    extract_r_like(x, call)
-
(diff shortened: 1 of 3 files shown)


R/generics.R
```diff
@@ -68,25 +68,12 @@ print.tensorflow.python.ops.variables.Variable <- print.tensorflow.python.framew
   check_zero_based(call)
 
   r_like_extract <- getOption("tensorflow.r_like_extract", TRUE)
```
goldingn: It might be worth changing this back to `tensorflow.zero_based_extract` or similar.

goldingn: otherwise, exactly what I had in mind & looks great!

jjallaire:
Thanks for the feedback, I've changed the option to `tensorflow.one_based_extract = TRUE`. If you give me the thumbs up I'll go ahead and merge.

goldingn:
:+1:

PR #206, opened 2017-11-18
fix as.matrix.python.builtin.EagerTensor

nakamichiworks:
Fixed minor mistakes in `as.matrix.python.builtin.EagerTensor()`


jjallaire:
Thanks! I think I was one cup of coffee short of where I needed to be when I wrote that code :-\

PR #212, opened 2017-12-19
Add support for export_savedmodel() over a plain session and tensors

javierluraschi:
There is a lot of boilerplate we can simplify while using `export_savedmodel()` in plain TensorFlow, no use of `tfestimators` not `keras`, which enable us to save models with something like:

```
  sess <- tf$Session()

  # Train MNIST but return a list with the input/output pointers
  tensors <- train_mnist(sess) 

  temp_path <- tempfile()
  export_savedmodel(
    sess,
    temp_path,
    inputs = list(images = tensors$input),
    outputs = list(scores = tensors$output)
  )
```


R/save.R
```diff
@@ -46,3 +46,44 @@ view_savedmodel <- function(
 
   tensorboard(log_dir = log_dir)
 }
+
+#' @export
+export_savedmodel.tensorflow.python.client.session.Session <- function(
```
jjallaire: You can actually explicitly provide the `inputs` and `outputs` parameters in the method signature (implementations of S3 methods can specify additional arguments). Also, in order for it to be a valid S3 method implementation you need to name the core parameters the same (i.e. `sess` needs to be `object`). 

I also wonder if the object which we are dispatching over should be a graph rather than a session?

jjallaire:
We could also dispatch on an R list that includes 2 named elements (inputs and outputs). I believe we can still determine the session by calling methods on the tensors but I'm not 100% positive.

javierluraschi:
Right, I thought of the list as well but it's less discoverable. The graph could also make sense but we use the `sess` in the export, so it seems more appropriate to use `sess`. Nice about having additional parameters, changed.

PR #213, opened 2017-12-20
Use correct export signature under export_savedmodel()

javierluraschi:
`export_savedmodel()` is saving using the `signature_def_map_class_dig` signature as opposed to `serving_default` which is more appropriate and required in some cases.

### DIFF
--- a/R/save.R
+++ b/R/save.R
@@ -64,9 +64,8 @@ export_savedmodel.tensorflow.python.client.session.Session <- function(
     method_name = tf$saved_model$signature_constants$PREDICT_METHOD_NAME)
 
   signature_def_map_class_dig <- tf$saved_model$signature_constants$DEFAULT_SERVING_SIGNATURE_DEF_KEY
-  signature <- list(
-    signature_def_map_class_dig = prediction_signature
-  )
+  signature <- list()
+  signature[[signature_def_map_class_dig]] <- prediction_signature
 
   builder <- tf$saved_model$builder$SavedModelBuilder(export_dir_base)
 


R/save.R
```diff
@@ -64,9 +64,8 @@ export_savedmodel.tensorflow.python.client.session.Session <- function(
     method_name = tf$saved_model$signature_constants$PREDICT_METHOD_NAME)
 
   signature_def_map_class_dig <- tf$saved_model$signature_constants$DEFAULT_SERVING_SIGNATURE_DEF_KEY
-  signature <- list(
-    signature_def_map_class_dig = prediction_signature
-  )
+  signature <- list()
+  signature[[signature_def_map_class_dig]] <- prediction_signature
```
jjallaire: I'm sure this is all correct but I don't see exactly how this changes things to serving_default?

jjallaire:
Do we need the same change in tfestimators and keras?

R/save.R
javierluraschi: I was expecting this:

```r
a <- "foo"
list(a = 10)
```

to be interpreted as:

```
$foo
[1] 10
```

but instead produces:

```
$a
[1] 10
```

So it effectively changes the name of the signature of the SavedModel. Maybe there is a better way to accomplish this in R? But regardless, this approach would work.

javierluraschi:
Thanks for asking, but we are good in the other packages:

https://github.com/rstudio/keras/blob/474bec5fe5bec47046c7d7a3ebf3a8bde610f436/R/model-persistence.R#L281-L284

https://github.com/rstudio/tfestimators/blob/b63e760c6a98957ceb7906f750a3a903397da5a5/R/tf_estimator.R#L357

PR #216, opened 2018-01-11
Additional params and naming for export_savedmodel()

javierluraschi:
See https://github.com/rstudio/tfdeploy/issues/2. Validated that this change does not break `keras` not `tfestimators` since those functions still handle the old parameter names; however, using the latest `tensorflow` package without upgrading `keras` or `tfestimators` would break this; however, will send PR to each package to depend on devel.

### DIFF
--- a/R/estimator-generics.R
+++ b/R/estimator-generics.R
@@ -62,7 +62,7 @@ train_and_evaluate <- function(object, ...) {
 #' [tfestimators][tfestimators::export_savedmodel.tf_estimator()] packages.
 #'
 #' @param object An \R object.
-#' @param export_dir A string containing a directory in which to export the
+#' @param export_dir_base A string containing a directory in which to export the
 #'   SavedModel.
 #' @param overwrite Should the \code{export_dir} directory be overwritten?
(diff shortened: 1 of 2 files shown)


R/estimator-generics.R
```diff
@@ -75,7 +77,12 @@ train_and_evaluate <- function(object, ...) {
 #'
 #'
 #' @export
-export_savedmodel <- function(object, export_dir_base, ...) {
+export_savedmodel <- function(
+  object,
+  export_dir,
+  overwrite = TRUE,
+  versioned = FALSE,
```
jjallaire: I wonder if this should default to `!overwrite`

jjallaire: Don't we need to also update the `export_savedmodel` for TF session elsewhere in the package?

jjallaire:
Do we still need to implement `overwrite` and `versioned` for saving tensorflow models based on the current session and input/output tensors? (e.g. here https://github.com/rstudio/tensorflow/blob/9d8edba5dacf645805d1dcb3f1f009c939b38cbf/R/save.R#L51)

PR #219, opened 2018-01-15
Add support to export SavedModels as text

javierluraschi:
Add support to export SavedModels models as text. In some corner cases, binary distribution of files might be blocked, say CRAN, attachments, etc. I don't think this change is a must for the upcoming TensorFlow CRAN release.

See also: https://github.com/rstudio/keras/pull/253, support in `tfestimators` is already enabled.


jjallaire:
Let's wait until after the CRAN release.

Do we need this for tfestimators as well?

javierluraschi:
It's already implemented in tfestimators, but was missing in keras and tensorflow.

Issue #230, opened 2018-04-02
unexpected behavior in `[` strided_slice   


t-kalinowski:
missing elements in an index seem to be ignored.

``` r
library(tensorflow)
d <- c(2, 4, 2)
X <- array_reshape(1:prod(d), d)

x <- tf$convert_to_tensor_or_sparse_tensor(X)
sess <- tf$Session()
x_1_3 <-   sess$run(x[1,c(1,3),])
x_1_2_3 <- sess$run(x[1,c(1,2,3),])
identical(x_1_3, x_1_2_3) # should be FALSE
#> [1] TRUE

# expected output:
X[1, c(1,3),]
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]    5    6

# actual output:
x_1_3
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]    3    4
#> [3,]    5    6
```

Created on 2018-04-02 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

jjallaire:
@goldingn Is the above expected behavior?

bwlewis:
`[`is eventually mapped in this example to `tf.strided_slice` (https://www.tensorflow.org/api_docs/python/tf/strided_slice)   here: https://github.com/rstudio/tensorflow/blob/master/R/extraction.R#L104

Ye gads! `tf.strided_slice` has an astoundingly horrible API and it does not look like it can handle generic indexing. It hurts my brain to read through that stuff!  It's not clear from some quick and dirty experiments I just ran that the suggested https://www.tensorflow.org/api_docs/python/tf/Tensor#__getitem__  method is much better. (I was not able to replicate generic numpy indexing with that either).

Since the tensorflow APIs are so awful, it might be best to just throw an error when presented with non-contiguous bracket function subset ranges until tensorflow itself is improved. Just a cheap proposal...

See also several open/gripe issues on tensorflow related to this like https://github.com/tensorflow/tensorflow/issues/4638 and   https://github.com/tensorflow/tensorflow/issues/206

For complete reference, the actual low-level tensorflow implementation is here:  https://github.com/tensorflow/tensorflow/blob/7a0def60d45c1841a4e79a0ddf6aa9d50bf551ac/tensorflow/core/ops/array_ops.cc#L1501

jjallaire:
Wow, that's really disappointing that TF isn't smarter about this.

[...]

PR #234, opened 2018-04-17
revamp `[.tensorflow.tensor`

t-kalinowski:
This is a new implementation of `[.tensor`, as discussed in issue #230.
I believe with this PR, all the functionality available from python brackets when indexing a tensor is now available in R as well.

New user facing features:

+ python-style strided slices. Accepted syntax is:
```
X[start:stop:step]
x[`::step`]
X[NULL:NULL:step]                    # missing args can alternatively be NULL
x[seq(start, stop, by = step)]       # standard R sequence translated to a py_slice()
```

+ indexing with tensors is supported
```
X[tf$constant(1L)]
X[tf$contstant(1L):tf$constant(5L)]
```

+ negative-index's are interpreted python-style. Only scalar negative integers are valid (scalars passed directly or within a py_slice(). In standard R sequences as produced by `seq()`, negative numbers produce an error)
```
x[-1]
x[-1:NULL]
```
On first use in an R session, the user is warned that negative numbers are
interpreted differently from standard R rules ( this warning can be turned by
setting the appropriate `options()`)

+ `py_ellipsis`, expands to the number of dimensions required
syntax is
```
x[1, py_ellipsis()]
```

+ `tf$newaxis` (alias for `None` / `NULL`) creates a new dimension. Syntax is:
```
x[1, tf$newaxis]
x[1, NULL]
```

Breaking changes:
+ passing a zero to `[` when `options(tensorflow.one_based_extract)` is `TRUE`
or `NULL` now throws an error instead of a warning.

Discussion points / questions:
+ Most of this logic is generalizable to numpy arrays, should parts of this be
ported over to reticulate?

+ Regardless of the previous question, should py_ellipsis() and/or py_slice() live in reticulate?

+ I'm concerned if the python-style interpretation of negative-indexes is going
to be confusing. Is a warning the right way to go here, or is there another suggestion?


goldingn:
Awesome work @t-kalinowski!
I'm slammed today, but will try to review in the next couple of days

t-kalinowski:
Thanks @bwlewis and @goldingn ( I look forward to any more comments you have)

I think overall I'm pretty happy with how this turned out. One thing that I'm not quite happy with here is how global `options()` are being used. It: 
   1) feels cluttered because there are so many (long-named) options (`tensorflow.one_based_extract`, `tensorflow.warn_negative_extract_is_python_style`, `tensorflow.py_slice_stop_is_inclusive`, `tensorflow.warn_tensors_in_extract_passed_as_is` (and possibly one more based on @bwlewis's comment `tensorflow.check_idx_out_of_bounds`) and 
    2) Is difficult to use in a package because the global `option()` is the ONLY way these options can be specified

I'm considering adding explicit args to the formals, something like:  
```
[.tensor(x, ..., drop = TRUE, one_based = getOption("tensorflow.one_based_extract"), inclusive_stop = getOption("tensorflow.py_slice_stop_is_inclusive), etc...)
```
or perhaps a list of options and a convenience options constructor, for something like 
```
[.tensor(x, ..., drop = TRUE, opts = tf_extract_opts())
```
where tf_extract_opts is just a list constructor, similar to kind used in the in leaflet package. 
```
tf_extract_opts <- function(
    one_based = getOption("tensorflow.one_based_extract"), 
    inclusive_stop = getOption("tensorflow.py_slice_stop_is_inclusive), 
    etc...) {
 list(environment()
}
```

and then package authors could define a package opts, and always refer to that

```
# within a package
OPTS <- tf_extract_opts(...)

#' @export
fun_foo <- function(...) { 
   x[i,j, opts = OPTS]  # where x is a tensor
}
```


This would address concern b), by allowing package authors to specify these options for each specific invocation of `[`, but I don't think it fully addresses concern a) (about clutter). 

Do you guys have any other suggestions about how to address this?

jjallaire:
I like your idea about having a single `options` argument to the extract operator that is populated by a default list.

Will wait for @goldingn to weigh in before merging this PR.

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+  start <- x[1L]
+  end <- x[length(x)]
+  step <- x[2L] - start
+
+  if(is.array(x))
+    stop("Subsetting tensors with R matrixs or arrays is not currently supported")
+
+  if (start < 0L || end < 0L ||
+      any(suppressWarnings(x != seq.int(from = start, to = end, by = step))))
+    stop("only positive sequences with a single step size are accepted\n",
```
goldingn: the error and warning messages throughout would be a little easier to read without the newline `\n` here. E.g. it rendered for me as:
```
 Error in R_slicing_sequence_to_py_slice(x, one_based) : 
  only positive sequences with a single step size are
accepted
if indexing with sequences not specified by a call to `:`, 
```

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+         "if indexing with sequences not specified by a call to `:`,\n")
```
goldingn: remove comma `,`

consider add `call. = FALSE` for consistency with elsewhere in the package (though it isn't used everywhere)

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+    FALSE
+    # should return false if the colon is not the top level call, e.g. -(1:4)
+}
+
+
+should_warn_about_negative_indices <- (function() {
+  user_was_warned <- FALSE
+
+  function(x) {
+    if (user_was_warned)
```
goldingn: If these options are all bundled into one, it might be tidier to define a single environment in the namespace in which to cache whether the user was warned, rather than setting up these constructor functions. E.g. something like:

```r
warned_about <- new.env()
warned_about$negative_indices <- FALSE
warned_about$something_else <- FALSE

should_warn_about_negative_indices <- function () {
  if (warned_about$negative_indices) {
    FALSE
  } else if (missing(x)) {
    !isFALSE(getOption("tensorflow.warn_negative_extract_is_python_style"))
  } else {
    warned_about$negative_indices <- TRUE
  }
}
```

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+  else
+    stop("0 is not a valid slice spec if `one_based_extract` is TRUE")
+
+}
+
+
+# ----- py_* functions ----
+
+# should this live in reticulate?
+#' @export
```
goldingn: This will need to be documented (whichever package it goes in).

At the moment, it causes this warning on checking the package:
```
checking for missing documentation entries ... WARNING
Undocumented code objects:
  ‘py_ellipsis’
All user-level objects in a package should have documentation entries.
See chapter ‘Writing R documentation files’ in the ‘Writing R
Extensions’ manual.
```

R/extract.R
```diff
@@ -0,0 +1,338 @@
+
+#' @export
+"[.tensorflow.tensor" <-
```
goldingn: It's a shame (though I don't think important) that the signature for this function is different than in base. `[` isn't an S3 generic, so there's no warning on package checking, so I think we're OK to do this. Handling the `i, j, ..., drop` signature in my previous attempt caused all sorts of pain.

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+
+should_warn_about_negative_indices <- (function() {
+  user_was_warned <- FALSE
+
+  function(x) {
+    if (user_was_warned)
+      FALSE
+    else if (missing(x))
+      !identical(
+        getOption("tensorflow.warn_negative_extract_is_python_style"), FALSE)
```
goldingn: you can use the `isFALSE()` function you defined below

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+
+
+translate_one_based_to_zero_based <- function(x) {
+  stopifnot(is_scalar_integerish(x))
+  if (x > 0L)
+    x - 1L
+  else if (x < 0L)
+    x
+  else
+    stop("0 is not a valid slice spec if `one_based_extract` is TRUE")
```
goldingn: maybe reword to: "cannot use 0 in slice syntax if `one_based_extract` is TRUE"
since `slice spec` might be confusing

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+    step <- (step %||% 1L) * -1L
+
+  # python slice() is exclusive of stop, but R is typically inclusive of all
+  # elements in a slicing sequence. Default to R behavior.
+  if (isTRUE(inclusive_stop)) {
+    if(!is.null(stop))
+      stop <- stop + as.integer(sign(stop) * sign(step %||% 1L))
+  }
+
+  import_builtins()$slice(start, stop, step)
```
goldingn: there's an overhead (albeit slight) to importing builtins on every call. Could this slice op be defined above instead, like:
```r
builtin_slice <- import_builtins()$slice
```
Then used here as:
```r
builtin_slice(start, stop, step)
```

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+
+}
+
+
+# ----- py_* functions ----
+
+# should this live in reticulate?
+#' @export
+py_ellipsis <- function()
+  import_builtins(FALSE)$Ellipsis
```
goldingn: could save some overhead on loading builtins by doing:

```r
py_ellipsis_internal <- import_builtins(FALSE)$Ellipsis
py_ellipsis <- function () py_ellipsis_internal
```

R/extract.R
```diff
@@ -0,0 +1,338 @@
...
+                 .friendly_description),
+           call. = FALSE)
+  }
+  invisible(TRUE)
+}
+
+check_is_TRUE_or_FALSE <- function(...) {
+  check_vars(...,
+             .fun = function(x)
+               identical(x, TRUE) || identical(x, FALSE),
```
goldingn: could use `isTRUE` and `isFALSE` here

t-kalinowski:
OK, thanks @jjallaire, @goldingn, I will try to push to this branch this evening (or maybe tomorrow evening), adding the argument `options = tf_extract_opts()`

R/extract.R
t-kalinowski: @jjallaire, do you know if the missing `i, j,` args might cause an issue somewhere?

R/extract.R
t-kalinowski: Are there going to be any issues with null xptrs if we do this? I haven't explored the internals of reticulate really, but I think that python might not always be available when the package code is evaluated, or if it is, the python session used then might be a different python session than the one that will be available when the user calls `py_ellipsis()`. 

Maybe we need to do something in `.onLoad()` if we want to cache a pointer to builtins?

jjallaire: Definitely don't do anything in onLoad (as this might interfere with binding to the environment where TF is). This code should be fine as it is (loading builtins on demand).

goldingn: Good points. The overhead is minimal, so I reckon just ignore these two comments of mine

t-kalinowski: followup: one way we could acheive this with minimal complications:
```R
.py_ellipsis <- NULL
py_ellipsis <- function() {
  .py_ellipsis %||% (import_builtins(FALSE)$Ellipsis ->> .py_ellipsis)
}
```
ditto for `py_slice()`

jjallaire: I could be wrong but I thought that `.py_ellipses` would be sealed after package load? (typically for package level mutables I put those in the `.globals` environment).

t-kalinowski: You are correct, I just gave it a try and got this error message
```
Error in .py_ellipsis <<- import_builtins(FALSE)$Ellipsis : 
  cannot change value of locked binding for '.py_ellipsis'
```

Something like this seems to work:
``` R
.globals <- new.env(parent = emptyenv())
py_ellipsis <- function() {
  .globals$py_ellipsis %||%
    (.globals$py_ellipsis <- import_builtins(FALSE)$Ellipsis)
}

```

jjallaire: There is already a `.globals` in the TF package that you'll want to make use of instead: https://github.com/rstudio/tensorflow/blob/master/R/package.R#L26

t-kalinowski:
I've incorporating Nick's feedback and the `options` idea, added some docs and examples, and took the opportunity to refactor some parts of the code. In total, the changes ended up being a little more substantial than I initially thought they were going to be. Another full code review is probably appropriate. 

There is a new `options` argument to `[`, and a companion `tf_extract_opts()` function. `tf_extract_opts()` offers some fine-grained controls for what translations and warnings are issued, and one convenience argument `style` for changing all the arguments at once. Now we have a few ways to specify options: 

``` R
x[1, style = "R"]  # style can be "R" or "python" 
options(tensorflow.extract.style = "R")
opts <- tf_extract_opts("python")
x[1, options = opts]
```
Specifying a `style` by default also automatically silences the warnings.

After this refactoring I think it would be pretty clean to also translate tensors from R to python (I actually started down that path a bit). But I decided it wasn't a good idea. If there are any thoughts about that, please let me know.

t-kalinowski:
Also, I incorporated @bwlewis's suggestion about out-of-bounds check

t-kalinowski:
So, I don't understand why the travis build is failing. When I run rcmd check on my computer with the same options (--as-cran --no-manual) I don't get that error. The log says that it fails after calling `str` (that is, tries to print) one of the examples in `tf_extract_opts()`. But that example runs fine on my computer. In short, I can't reproduce the issue, so I'm not quite sure how to fix it. Any help would be appreciated.

t-kalinowski:
I marked the offending example as `donrun{}` for now to make the build pass. Also, I renamed one of the arguments in tf_extract_opts() to address a rcmdcheck note about truncation in the pdf manual.

goldingn:
Wow, loads of work!

Not quite sure what you meant by this:

> After this refactoring I think it would be pretty clean to also translate tensors from R to python 

I'll try to take a look at the code on Monday

t-kalinowski:
Thanks @goldingn....  I meant that 
```
x[tf$constant(1L):tf$constant(2L)] 
```
could be translated such that 1 based -> 0 based, inclusive stop -> exclusive stop.

goldingn:
Ah right, thanks! Yeah, that seems like an edge case that might confuse things. I agree it would be best to leave that functionality out.

jjallaire:
@goldingn I'm wondering what our test/integration strategy should be here. I realize this is an ambitious change and I certainly don't want to break anyone (but realize that it's not out of the question that there is some breakage). Are we okay to merge this now and then leave it on master for 2 or 3 weeks before submitting to CRAN or should there be some additional testing?

@t-kalinowski I realize you don't have access to the repo where we document TF but could you suggest some enhancements to these docs to along with the PR: https://tensorflow.rstudio.com/tensorflow/articles/using_tensorflow_api.html#tensor-extraction

goldingn:
@jjallaire as far as I can tell, these changes *should* be backwards compatible with the previous implementation (right, @t-kalinowski?) There were quite a few tests on the previous implementation, and they all still pass, which is good news.

I think leaving it on master for a bit is a good idea. I also think we should check the reverse dependencies on CRAN too. Of those dependencies, 3 are RStudio's: keras, tfdatasets, tfestimators, 1 is mine: greta, and 3 are other people's: sgmcmc, kerasformula, nimble.

I've just checked this against greta's current release version (0.2.4) and it flagged some issues. I'll try to pin those down now, and review the recent edits.

I'll try to run reverse dependency checks for the other packages too.

goldingn:
@t-kalinowski the issue I'm having with greta is that there's one internal method that selects an element of a 2D tensor with a single dimension. That was permitted in the previous tensorflow, but not in this version:
```r
x <- matrix(1:3)
x2 <- tf$constant(x)
x[1]
x2[1]
```
> ```
>  Error in `[.tensorflow.tensor`(x2, 1) : 
>   Incorrect number of dimensions supplied. The number of supplied arguments, (not counting any
>   NULL, tf$newaxis or np$newaxis) must match thenumber of dimensions in the tensor, unless a
>   py_ellipsis() was supplied 
> ```

The behaviour in the current version of tensorflow matches Python tensorflow, subsetting on the first dimension of the tensor. E.g.:

```r
x <- matrix(1:6, ncol = 2)
x2 <- tf$constant(x)
dim(x2[1])
```
> ```
> [1] 2
> ```

How difficult do you think it would be to implement this previous behaviour?

NAMESPACE
```diff
@@ -63,11 +63,13 @@ export(iterate)
 export(np_array)
 export(parse_arguments)
 export(parse_flags)
+export(py_ellipsis)
```
goldingn: @jjallaire how would you feel about `py_ellipsis` being exported in reticulate, rather than here?

R/extract.R
```diff
@@ -0,0 +1,604 @@
...
+#' @param x Tensorflow tensor
+#' @param ... slicing specs. See examples and details.
+#' @param drop whether to drop scalar dimensions
+#' @param style One of `"python"` or `"R"`.
+#' @param options An object returned by `tf_extract_opts()`
+#'
+#' @export
+#'
+#' @examples
+#' grab <- (function() {
```
goldingn: I'm not sure about putting this grab function and overwriting `print.tensorflow.tensor()` in the examples, it would probably be clearer to readers of the helpfile to define `sess <- tf$Session()` at the top, then do `sess$run()` in place of `grab()`.

R/extract.R
```diff
@@ -0,0 +1,604 @@
...
+
+    x$`__getitem__`(dots)
+  }
+
+
+
+#' Tensor extract options
+#'
+#' @param style one of `NULL` (the default) `"R"` or `"python"`. If supplied,
+#'   this overrides all other options. `"python"` is equivelant to all the other
```
goldingn: equivelant -> equivalent

R/extract.R
```diff
@@ -0,0 +1,604 @@
...
+#'   arguments being `FALSE`. `"R"` is equivelant to
+#'   `warn_tensors_passed_asis` and `warn_negatives_pythonic`
+#'   set to `FALSE`
+#' @param ... ignored
+#' @param one_based TRUE or FALSE, if one-based indexing should be used
+#' @param inclusive_stop TRUE or FALSE, if slices like `start:stop` should be
+#'   inclusive of `stop`
+#' @param disallow_out_of_bounds TRUE or FALSE, whether check are performed on
```
goldingn: check -> checks, and the sentence seems to be unfinished?

R/extract.R
```diff
@@ -0,0 +1,604 @@
...
+
+
+translate_one_based_to_zero_based <- function(x) {
+  stopifnot(is_scalar_integerish(x))
+  if (x > 0L)
+    x - 1L
+  else if (x < 0L)
+    x
+  else
+    stop("cannot use 0 in slice syntax if `one_based` is TRUE")
```
goldingn: yeah, that's much clearer!

R/extract.R
```diff
@@ -0,0 +1,604 @@
...
+      "options(tensorflow.extract.disallow_out_of_bounds = FALSE)"))
+}
+
+
+stop_if_any_zeros <- function(dots) {
+  recursivly_check_dots( dots,
+    function(d) isTRUE(any(d == 0)),
+    if_any_TRUE = stop( paste(
+        "It looks like you might be using 0-based indexing to extract using `[`.",
+        "The tensorflow package now uses 1-based (R-like) extraction by default.\n",
```
goldingn: maybe remove the phrase 'R-like', since the default is now different than the R-like syntax?

R/extract.R
```diff
@@ -0,0 +1,604 @@
...
+}
+
+warn_if_any_tensors <- function(dots) {
+  recursivly_check_dots(dots,
+    function(x) TRUE, classes = "tensorflow.tensor",
+    if_any_TRUE = {
+      warning(call. = FALSE,
+        "Indexing tensors are passed as-is to python, no index offsetting or ",
+        "R to python translation is performed. Selected options for one_based ",
+        "and inclusive_stop are ignored and treated as FALSE to silence this warning, set ",
```
goldingn: "FALSE to silence" -> "FALSE. To silence"

R/extract.R
jjallaire: Also, all of the examples that call TF need to be marked with `\dontrun`

NAMESPACE
jjallaire: I think I would rather have this function be renamed to something more fitting for it's usage here (e.g. `all_dims()`, no idea if that's the best name but I think something that invokes it's purpose rather than the inside baseball of `py_ellipsis()` would be better.

goldingn:
I haven't done the revdep checks, since they take a long time and I know that one issue with greta is outstanding. Once we've resolved that, I or anyone else can run them with `devtools::revdep()`.

t-kalinowski:
Thanks @goldingn for the review. re supplying a single arg to a multidimentional object, what is the expected value of `x2` in the two examples you provide?

Tensorflow (and python) treat the arg as if it was supplied to the first dimension and implicitly supplies a py_ellipsis() at the end to fill in the rest. R on the other hand pretends the dim attribute doesn't exist and just selects the appropriate element(s) from the underlying vector. so, with
```
x <- matrix(1:6, ncol = 2)
```
in python `x[2]` returns the second row `c(2, 5)`, and in R it returns the second element in the vector `2`.

Getting the python behavior is as simple as removing the appropriate check. Getting the R behavior is going to require a little more work.

Is the python behavior the one you want?

NAMESPACE
t-kalinowski: I'll try to think of another name. `all_dim()` doesn't feel quite right (it's for getting the remainder of the dims, not _all_ the dims).
first two that come to mind: `expand_commas()` and `rest_of_dims()

goldingn:
Yes, the current version of tensorflow has the python behaviour (extract on the first dimension), so that's what's needed to clear that backward incompatibility issue.

Having the R approach as an option would be nice, but not so important.

R/extract.R
t-kalinowski: @goldingn, are you sure? I feel like it demonstrates some neat tricks that help make interactive experimentation at the console much easier and seem harmless enough, but I admittedly don't have much exposure to how most people use R.

goldingn: Explicit use of `sess$run()` is what's done throughout the rest of the examples, and in the tensorflow docs users are pointed to, so I think it makes sense to be consistent to reduce cognitive overhead. It's JJ's call though.

Showing what the outputs are is nice though. That `grab()` function I wrote for the tests is now superseded by [TF eager mode](https://www.rdocumentation.org/packages/tensorflow/versions/1.5/topics/tfe_enable_eager_execution), so you could maybe use that instead?

t-kalinowski:
Re subsetting only on the first dimension of a multidimensional object: I'm currently thinking that requiring an explicit `py_ellipsis()` (or whatever we rename it to) is the right way to go for the R interface. 

The implicit trailing `py_ellipsis` in `x[1]` is going to be surprising and unexpected to R users. However, the fact the pythonic usage style is already in use on CRAN in greta gives me pause. Should we just accept this as part of the feature-list and document this pythonic behavior in the examples (the fact that `x[1]` is equivalent to `x[1, py_ellipsis()]`, or should we plan to eventually require an explicit py_elipsis() and deprecate the current pythonic usage (e.g, throw a warning now, eventually in a future update make it an error)

goldingn:
Yeah, it's a tricky one - there are lots of awkward decisions to make in bridging these two languages! When I first started thinking about this, my approach was to implement just the bits of subsetting syntax that were common to both Python and R. That's conceptually clean but pretty limiting.

I think that ultimately moving to the more explicit indexing; requiring a `py_ellipsis`, will be safer and easier for users to understand. A deprecation period is a good idea. If greta's the only CRAN package that change will break, then I can patch it easily, though I haven't checked the impact on the other packages yet though (and unfortunately won't have any more time this week).

NAMESPACE
t-kalinowski: re renaming py_ellipsis():
I'm currently thinking about `expand_commas()` or `all_other_dims()`, slightly leaning towards `expand_commas()`. Here's what it might look like in usage:
``` R
x[1, expand_commas()]
x[1, all_other_dims()]
```
@jjallaire, thoughts?

t-kalinowski: ok, I'll rename it to `all_dims()`.

t-kalinowski:
Ok, I'm implemented the suggested changes and as far as I can tell this is good to go.

I ran into some issues when checking revdep_check() and greta, but I don't think they're related to this PR. I got the same failed tests when running greta's tests using CRAN's tensorflow (testing greta both from CRAN and from the greta-dev/greta). Perhaps it's related to tf 1.8? I admittedly didn't dig into it too much beyond confirming that the error wasn't caused by `[` (as far as I can tell, `[` is not called in the failing tests.)

goldingn:
Yeah, that makes sense. The interface/behaviour of `tf.contrib.distributions` has changed, which is what's cause the CRAN and github releases of greta to fail tests with recent versions.

Installing your fork: `devtools::install_github("t-kalinowski/tensorflow@master")`;
tensorflow 1.6: `tensorflow::install_tensorflow(version = "1.6")`;
and testing with the latest greta release: `devtools::install_github("greta-dev/greta")`,
the tests all pass without error.

Some throw "Incorrect number of dimensions supplied." warnings saying that the old behaviour is now deprecated, which is as expected and easy for me to clear up.

R/extract.R
```diff
@@ -102,10 +102,14 @@
...
     if (length(dropNULL(dots)) != py_len(x$shape)) {  # NULL == tf$newaxis
       is_ellipsis <- vapply(dots, is_py_ellipsis, FALSE)
-      if (!any(is_ellipsis)) stop(
+      if(length(dropNULL(dots)) > py_len(x$shape))
+        stop( "Incorrect number of dimensions supplied. ", py_len(x$shape),
+          " required, but received ", length(dropNULL(dots)))
+      else if (!any(is_ellipsis)) warning(
         "Incorrect number of dimensions supplied. The number of supplied arguments, ",
         "(not counting any NULL, tf$newaxis or np$newaxis) must match the",
         "number of dimensions in the tensor, unless an all_dims() was supplied",
```
goldingn: "...all_dims() was supplied(this will..." -> "...all_dims() was supplied (this will..."
(add a space before parenthesis)

goldingn:
Just the one text edit there. Looks great otherwise!

t-kalinowski:
Thanks @goldingn!

jjallaire:
I think this is good to merge and I'd like to merge it soon so that we can integrated it into the next CRAN release. The only thing missing from my point of view is a writeup of the new behavior (could be mostly derived from the help) that would replace this within the existing documentation: https://tensorflow.rstudio.com/tensorflow/articles/using_tensorflow_api.html#tensor-extraction

@t-kalinowski The repo where those docs are is private, but if you could add suggested docs either as a comment to this issue or as a gist I can incorporate them.

t-kalinowski:
Ok, I'll write something up tomorrow evening (I doubt I'll be able to get to it tonight.)

jjallaire:
Am going to merge now. @t-kalinowski Let me know once you've got a draft writeup for the website and I'll post it there.

t-kalinowski:
will do! I have a draft started, but given the way the rest of the week is looking I won't be able to wrap it up until the weekend.

Glad it's finally merged!

t-kalinowski:
@jjallaire,

Here is some documentation you requested. If you think it's a bit too long for that context, then feel free to trim this down and just refer people to the ``` ?`[.tensorflow.tensor` ```


## Tensor Extraction

Tensor elements can be extracted either by using functions like `tf$gather()` and `tf$slice()`, or by using `[` syntax.

Extracting tensor elements with `[` in R is similar to extracting elements from standard R arrays, albeit with some minor differences in capabilities. In contrast to most `tf$` functions, `[` defaults to R style 1-based rather than 0-based indexes. Currently, only numeric indexes in `[` are supported (no logical or character indexes)

Extracting works identically to R arrays if the slicing index is missing, supplied as a scalar, or as a sequence (e.g, created by `:` or `seq_len()`)  
```R
x[,1] )      # all rows, first column
x[1:2,] )    # first two rows, all columns
x[,1, drop = FALSE] ) # all rows, first column, but preserving the tensor rank
```

`[` also supports slices with a strided step, which can be specified in traditional R style with `seq()` or with a python-style second colon. If you are unfamiliar with python-style strided step syntax, see here for a [quick primer](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html#basic-slicing-and-indexing)
```R
x[, seq(1, 5, by = 2)]  # R style
x[, 1:5:2]              # Equivalent python-style strided step 
```

Missing arguments for python syntax are valid, but they must be supplied as NULL or whole expression must by backticked.
``` R
x[, `::2`] 
x[, NULL:NULL:2] 
x[, `2:`] 
```

`[` also accepts  `tf$newaxis` and `all_dims()` as arguments
```R
x[,, tf$newaxis]
x[all_dims(), 1] # all_dims expands to the shape of the tensor
```

An important differences between extracting R arrays and tensorflow tensors with `[` is how negative numbers are interpreted. For tensorflow tensors, negative numbers are interpreted as selecting elements by counting from tail (e.g, they are interpreted python-style).  

```R
x[-1,]  # the last row
```

Tensors are accepted by `[` as well, but note that tensors suplied to `[` are not translated from R to python. Meaning that tensors are interpreted as 0-based, and if slicing a range with `:`, then the returned arrays is exclusive of the upper bound.
```R
x[, tf$constant(1L)] # second column
x[, tf$constant(0L):tf$constant(2L)] # first two columns
```

If you are translating existing python code to R, note that you can set an option to have all `[` arguments be interpreted pure-python style by setting `options(tensorflow.extract.style = "python")`. See ``` ?`[.tensorflow.tensor` ``` for additional options and details.

jjallaire:
Great! I just published those docs on the website.

And thanks again for your incredible work here :-)

PR #271, opened 2018-10-13
python_unix_bin using TENSORFLOW_PYTHON

zengfengbo:


skeydan:
I think this function is intended to determine the location of Python as provided by the OS.
It cannot take into account all possible environment variables (naming conventions) users might think of to manage their PATHs...

jjallaire:
The idea behind `python_unix_bin()` was to strongly prefer system installations of Python as we had all kinds of problems with custom builds that were on the PATH, homebrew builds, etc. Since we know that TF is well tested with system versions of Python it seemed preferable to use those in preference to whatever else might be installed. Note that we also check for python3 when python2 isn't found: https://github.com/rstudio/tensorflow/blob/10cefc322786afd562675184517da2bb73d5deeb/R/install.R#L102

Issue #279, opened 2018-10-31
Add Group "Summary" generics


t-kalinowski:
It would be nice if we also had tensor methods for the Group "Summary" functions `any, sum, prod, min, max, range`
They could all be simple wrappers like so: 
all  -->  tf.reduce_all() 
any -->  tf.reduce_any() 
sum --> tf.reduce_sum(), 
same for prod min, max 
range --> tf.concat([min(x), max(x)])

t-kalinowski:
Also, Group Complex generics would be nice
Arg, Conj, Im, Mod, Re

skeydan:
Hi, yeah, a PR would be great!

PR #280, opened 2018-11-18
generics for complex tensors

t-kalinowski:
This is partially addressing #279, implementing methods for the Group "Complex" generics. The complex methods I think are a net-win.

I did not include methods for the Group "Summary" generics in this PR because I had second thoughts about if implementing them is a good idea for these reasons:

+  the meaning of `...` would be different between the base R function and the method for tensors.
+  All of these functions are already neatly accessible from the keras backend functions `k_max k_min` etc..

My initial motivation for wanting Summary generics was because it would allow me to write one code path in functions that can take either tensors or R objects as arguments (without coercing everything to a tensor), but now I'm also second guessing if that's even a good idea, or if it will lead to code that's brittle and difficult to reason about. For the record, I was imagining something like this:

```R
#' @export
max.tensorflow.tensor <- function(x, ...) {
  tf$reduce_max(x, ...)
}

#' @export
min.tensorflow.tensor <- function(x, ...) {
  tf$reduce_min(x, ...)
}

#' @export
all.tensorflow.tensor <- function(x, ...) {
  tf$reduce_all(x, ...)
}

#' @export
any.tensorflow.tensor <- function(x, ...) {
  tf$reduce_any(x, ...)
}

#' @export
sum.tensorflow.tensor <- function(x, ...) {
  tf$reduce_sum(x, ...)
}

#' @export
prod.tensorflow.tensor <- function(x, ...) {
  tf$reduce_prod(x, ...)
}

#' @export
range.tensorflow.tensor <- function(x, ...) {
  tf$stack(list(min(x, ...), max(x, ...)))
}
```

Let me know what your thoughts are about implementing the Summary generics and I'll submit a separate PR if you think they're a good idea.


skeydan:
Hi, thanks for submitting this!

For the Summary generics, my totally personal and subjective view is that I like to keep R values and tensors as distinct as possible in my code (for example, because of different axis indexing), so I would probably not use them, but I don't know how others see this...

Footnote reg. axis indexing: of course it's not the objects per se but the difference in how we access them, so it's rather between (R) `sum` and `k_sum` on the one hand and `tf$reduce_sum` on the other, but anyway, for me it's easier if I have different code chunks in both cases.

t-kalinowski:
Yes, 1-vs-0 based axis indexing in this context is a can of worms. An alternative might be to completely avoid opening it by doing something like this
```r
max.tensorflow.tensor <- function(x) {
  tf$reduce_max(x)
}
```
Nothing is passed on into dots, and `max` therefore returns a scalar tensor, behaving similarly to the R side.

skeydan:
Yeah, right... I think I'm still wondering though (even with that way) if here we mightn't have some kind of trade-off: How much of use is it to advanced users vs. how confusing is it to the less experienced?

(Not sure though, just wondering.)

t-kalinowski:
Yeah, I was also ambivalent about the summary generics, that's why I didn't include them in the PR. However, complex generics however I think are pretty uncontroversial and I for one would use them plenty because they save quite a bit of typing.

skeydan:
Yeah, I agree!

PR #289, opened 2019-01-24
Support for autocomplete under TF 2.0

javierluraschi:
The logging API changed in TF 2.0, actually, a lot of things changed; this change enables autocomplete to help us troubleshoot further. @skeydan

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -105,19 +105,24 @@ tf_config <- function() {
   if (have_tensorflow) {
 
     # get version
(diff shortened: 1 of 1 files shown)


skeydan:
Hi @javierluraschi thanks for starting!

I see you've introduced 2 ways of deciding what to do:

(1) dependent on version
```
if (tf_v2()) {
```

and (2) dependent on module availability

```
if (reticulate::py_has_attr(tf, "version"))
```

When we continue making changes in this way, is there a heuristic which way is preferable (or preferred, aesthetically) in which context?

PR #291, opened 2019-02-01
wrapper function for accessing tf_probability

jjallaire:


skeydan:
works fine (provided tfp is installed, like we said, which we will take care of as soon as tf 1.13 is released)

PR #295, opened 2019-02-07
Support for use_compat() and finish TF 2.0 tests

javierluraschi:
Adding `use_compat()` convenience function to replace `tf` with `tf$compat$v1`. Notice that original `tf` object is retrieved for users that might want to access v1 in most cases but v2 in others.

Fixing a few most tests; however, the mnist examples require a full rewrite to use datasets since the `tf$contrib$learn$datasets` is now deprecated.


skeydan:
Hi Javier,

_use_compat_ is great to have! I'm not so sure about the examples though - 

```
if (tf_version() < "2.0") {
    # disable examples since tf_compat() requires session restart
    examples <- NULL
  }
```

Shouldn't we rather execute them with V1 only? 
For me, running the tests (including examples) with V2 now, I get errors not only from the examples but the updated for eager execution tests too (probably b/c `hello.R` has `use_compat("v1")`) . Somehow that command does not seem to work with the order I do things - restart R, set env variable, "test" - but still, has the consecutive non-examples tests failing... this might be easier to show you, later today...)


I think we could add a different set of examples for V2 instead (for example, `train_mnist_eager` already contains example code that runs with V2) ...

skeydan:
Strangely, I see that on travis, the `preview` test succeeds ... strange (as for me, locally, it doesn't).
However the `nightly` fails...

skeydan:
Rethinking what I wrote above, probably running the examples with V2 only, using compat, has the advantage that they still __get__ run, in the longer-term future ... let's just talk this through, later...

jjallaire:
This LGTM.

Note that for future reference another way to implement switching out the binding for an exported object is demonstrated here (via active bindings):

https://github.com/rstudio/reticulate/blob/master/R/zzz.R

I think it's better to leave the TF code as-is, but if for example we get feedback on the use of lock/unlock we could switch to this technique.

skeydan:
Looking good :-)

PR #301, opened 2019-02-15
install compatible version of tensorflow-probability

skeydan:
The requirements according to https://www.tensorflow.org/probability/install are:

- the current TFP release needs a "recent stable release of TensorFlow "
- `tfp-nightly` wants `tf-nightly` 

Testing with conda this yields

1.  when doing `install_tensorflow(version="1.12.0")`

```
(r-tensorflow) ] $ conda list | grep -e tf -e hub -e tb -e tens
# packages in environment at /home/key/anaconda3/envs/r-tensorflow:
tensorboard               1.12.2                    <pip>
tensorflow                1.12.0                    <pip>
tensorflow-hub            0.2.0                     <pip>
tensorflow-probability    0.5.0                     <pip>

```

and 

2. when doing `install_tensorflow(version="nightly")`

```
(r-tensorflow) $ conda list | grep -e tf -e hub -e tb -e tens
# packages in environment at /home/key/anaconda3/envs/r-tensorflow:
tb-nightly                1.13.0a20190214           <pip>
tensorflow-hub            0.2.0                     <pip>
tf-estimator-nightly      1.14.0.dev2019021501           <pip>
tf-nightly                1.13.0.dev20190215           <pip>
tfp-nightly               0.7.0.dev20190215           <pip>
```

Please let me know what you think.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -74,7 +74,7 @@ install_tensorflow <- function(method = c("auto", "virtualenv", "conda", "system
     extra_packages <- unique(c(extra_packages, "tfp-nightly"))
     extra_packages <- setdiff(extra_packages, "tensorflow-probability")
   }
-  if (substr(version, 1, 4) %in% c("1.12", "1.13")) {
+  if (version == "default" || substr(version, 1, 4) %in% c("1.12", "1.13")) {
     extra_packages <- unique(c(extra_packages, "tensorflow-probability"))
     extra_packages <- setdiff(extra_packages, ("tfp-nightly"))
   }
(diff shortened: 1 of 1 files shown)


R/install.R
```diff
@@ -71,12 +71,12 @@ install_tensorflow <- function(method = c("auto", "virtualenv", "conda", "system
   # extra packages
   extra_packages <- unique(c(extra_packages, c("keras", "tensorflow-hub")))
   if (version == "nightly") {
```
jjallaire: What about if the version is "default" ? (I think that will be typical once we remove the hack to force v1.12)

skeydan: Oh sorry, I must have been assuming we always set a version, but I see this once looked like 

```
if (identical(version, "default")) {

# default, no changes required
```

:-)

Fixed...

PR #305, opened 2019-02-27
install latest TF release 1.13.1

skeydan:
Strangely, there has never been a release `1.13.0`. 
I have to hard-code `1.13.1` because of this

```
    # workaround the fact that v1.X.1 releases often have no tarball
    version_split <- strsplit(version, ".", fixed = TRUE)[[1]]
    if (length(version_split) > 2 && version_split[[length(version_split)]] != "0")
      version_split <- c(version_split[-length(version_split)], "0")
```

Alternatively, we could check for available versions here, but I don't know if it's worth the overhead.

Still alternatively, I'm just thinking now as I write this - should we "try trusting them" ;-)) and remove the workaround completely? Right now we know there's a 1.13.1 with existing binaries, then it'll be some time till 2.0, and then we'll see?

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -268,7 +268,12 @@ install_tensorflow_conda <- function(conda, version, gpu, envname, packages, ext
 
   # create conda environment
   cat("Creating", envname, "conda environment for TensorFlow installation...\n")
-  python_packages <- "python=3.6" # as of TF v1.9 there are no python 3.7 binaries
+  if (substr(version, 1, 4) == "1.13") {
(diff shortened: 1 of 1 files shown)


jjallaire:
I've unfortunately gotten burned multiple times by "trusting them" to produce binaries for .1 releases. That said, this was more than a year ago so perhaps they've changed their protocols around releases by now. The problem is that it's really ugly when they don't produce a binary, as installations begin failing completely. So on balance I'd rather hard-code 1.13.1 than trust them.

Issue #191, opened 2017-10-26
R Tensorflow GPU running correctly?


zhongyifan92:
Hi, 

I just installed tensorflow GPU version in R, and wondering if there is some sample code to confirm that the computation is using GPU not CPU? 

Thank you!

zhongyifan92:
In addition,  I can monitor my GPU load to confirm that it works correctly. But just wondering if any sample code can show that in R?

bearloga:
Yes, see #171

PR #307, opened 2019-03-07
New feature: tf_gpu_configured function to check if GPU was correctly configured.

jcrodriguez1989:
Having configured my GPU I ran TensorFlow several times. However, when the Linux kernel was updated I noticed that sometimes the GPU was working no more, and I was realizing time later.
I thought it would be useful to have a function that reports the configuration status of the GPU.
I have also seen other people with the same intention of checking their GPU: [1](https://github.com/rstudio/tensorflow/issues/191), [2](https://stackoverflow.com/questions/38009682/how-to-tell-if-tensorflow-is-using-gpu-acceleration-from-inside-python-shell), [3](https://stackoverflow.com/questions/53221523/how-to-check-if-tensorflow-is-using-all-available-gpus), [4](https://stackoverflow.com/questions/51114771/how-to-ensure-tensorflow-is-using-the-gpu)

In this way I could add to my scripts something like:

``` r
if (!tf_gpu_configured(verbose = FALSE))
  stop("GPU not working!")
```

Also, I have checked that this function in the Python library exists at least since [version v1.0](https://github.com/tensorflow/docs/blob/r1.0/site/en/api_docs/python/tf/test/is_gpu_available.md)

This new function has as output depending on if the **GPU is configured**:

``` r
library("tensorflow")

gpu_config <- tf_gpu_configured()
```

    ## TensorFlow built with CUDA:  TRUE 
    ## GPU device name:  /device:GPU:0

``` r
gpu_config
```

    ## [1] TRUE

``` r
gpu_config <- tf_gpu_configured(verbose=FALSE)
gpu_config
```

    ## [1] TRUE

Or the **GPU is not configured**:

``` r
library("tensorflow")

gpu_config <- tf_gpu_configured()
```

    ## TensorFlow built with CUDA:  FALSE 
    ## GPU device name:

``` r
gpu_config
```

    ## [1] FALSE

``` r
gpu_config <- tf_gpu_configured(verbose=FALSE)
gpu_config
```

    ## [1] FALSE


jcrodriguez1989:
#191

mzeng-moodys:
the function returned NA for me
```
> library("tensorflow")
> gpu_config <- tf_gpu_configured()
Warning message:
Can not determine if GPU is configured. 
> gpu_config
[1] NA
```

skeydan:
@mzeng-moodys what do you get for 

```
tf$test$is_gpu_available()
```
?

PR #323, opened 2019-05-13
add tf_function

skeydan:
It's useful to have this even without autograph, as we can already wrap TF functions.

I've set the default of `autograph` to FALSE and added a note in the documentation, to make it clear that `autograph` as per today does not exist in R.


dfalbel:
Maybe we should error if `autograph = TRUE`?

skeydan:
makes sense, updated!

PR #333, opened 2019-06-19
Refactor install_tensorflow

dfalbel:
Delegates installation to `reticulate` functions.

I have set testing for different envs and OS's here: https://github.com/dfalbel/tfinstall

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -1,5 +1,3 @@
-
-
 #' Install TensorFlow and its dependencies
 #'
 #' @inheritParams reticulate::conda_list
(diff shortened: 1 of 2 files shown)


javierluraschi:
LGTM!

skeydan:
To me too, great! Just asking about https://github.com/rstudio/tensorflow/pull/333/commits/d71d12f1fd539d1dda729b4ef6d8c92c22b21f79, when I create a default new environment

```
conda create -n new python=3.7
```

I get these automatically:

```
-rw-rw-r--.  5 key key      132 Jun  9 22:30 libgcc_s.so
-rw-rw-r--.  5 key key   818984 Jun  9 22:30 libgcc_s.so.1

```

dfalbel:
You are right! I fixed this here: https://github.com/rstudio/tensorflow/pull/333/commits/4c58d7e8b24629963a9e9ee7660e2196fd98432a#diff-b5f7f6437b8e3c21cb2e9d6eb0ce1dd8L138

I added a conda_python_version argument to control which python version is installed in this env. Currently using 3.6.8 since 3.7 has some ABI incompatibility issues with tensorflow.

PR #338, opened 2019-06-27
also test against reticulate master

skeydan:


skeydan:
@dfalbel I think I don't have permissions to restart the build on appveyor, could you please do that? thanks

skeydan:
Merging because appveyor builds had failed because of `rlang`

PR #372, opened 2019-11-07
declare dependency on TensorFlow 2.0.0

kevinushey:
This allows the new machinery in `reticulate` to auto-install TensorFlow when the `tensorflow` package is loaded.

One question (whose resolution could be hairy): how do we decide between CPU vs. GPU installs of TensorFlow? Is it okay to just use the CPU version always?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,3 +48,9 @@ Imports:
 Roxygen: list(markdown = TRUE)
 Suggests: testthat, keras, tfestimators
 RoxygenNote: 6.1.1
+reticulate@R:
+  list(
+    packages = list(
+      list(package = "tensorflow", version = "2.0.0", pip = TRUE)
+    )
+  )


kevinushey:
Note: we can also remove the version requirement if it's not required here (ie if we are compatible with the latest version of TensorFlow)

skeydan:
The TF team is migrating to having the default (non-suffix) version include GPU support.
There is/will be a special cpu-only version, "for users concerned about package size".

AFAIK 2.0 does not implement that yet, but 1.15 and nightly do.
So "in the long run", installing the default version should be fine...

(see e.g.
https://groups.google.com/a/tensorflow.org/forum/#!topic/developers/iRCt5m4qUz0
https://www.tensorflow.org/install
)

Hm, writing this I'm starting to think this may actually be the cause for travis failing on `tfprobability` for versions 1.15 and nightly, but not 2.0! So we need to use the versions with `cpu` suffix on travis now...

cc @dfalbel

dfalbel:
Do you think it would make sense to do something like this:

```
reticulate@R:
  list(
    packages = list(
      list(package = Sys.getenv("TENSORFLOW_PACKAGE", unset = "tensorflow"), version = Sys.getenv("TENSORFLOW_VERSION", unset = "2.0.0"), pip = TRUE)
    )
  )
```

so users could easily override default package and version that get's installed?

dfalbel:
I just tested the workflow on Windows and Mac and it worked perfectly well. Thanks very much, Kevin!

kevinushey:
That seems reasonable. @jjallaire can you also sanity-check and share your thoughts?

jjallaire:
This seems fine to me. It's absolutely spectacular that there is no CPU/GPU distinction required. That day can't come soon enough! In the meantime, I think it's okay to remove the explicit version requirement as we do want to keep moving to the new version (unless you all think that there is so much breakage in the normal course of upgrades that we need to whitelist versions).

In terms of the environment variables, they seem fine and of course harmless when they aren't set. I wonder though if explicit installation workflows already available would be enough here?

dfalbel:
I think the only moment that the ENV vars would be important is in the case that we need to set the `version` argument and a user wants to use an older release. For example:

* I have TensorFlow 1.14 installed in my `r-reticulate` env and I want to keep with it.
* I install TensorFlow R package which sets `version=2.0.0` in the `reticulate@R` field because `2.1` breaks our package.

In this case, every time I load the TensorFlow R package the 2.0 version of the python package will be installed.

skeydan:
I'm not sure if I understand this correctly, so I'll just plainly ask :-)

Does this mean that on __package__ level, it is decided what version of the Python backend is installed?
In that case, how can a user have different (e.g.) TensorFlow versions at the same time?

I mean, right now this is easy, you just keep a bunch of virtualenvs resp. conda envs around...

kevinushey:
> Does this mean that on package level, it is decided what version of the Python backend is installed?

That's correct. The goal is to provide a workflow that works for users who don't want / need to care about Python, or managing their Python environments. That is, someone should be able to just call:

```
library(tensorflow)
tensorflow::model_all_the_things()
```

and it should "just work" for users. Power-users can override this with `reticulate` tools, e.g. `reticulate::use_virtualenv()` and friends, but for the average user who just wants to use the `tensorflow` R package without worrying about what Python machinery exists behind the scenes, we want to provide a workflow that can still work.

> In that case, how can a user have different (e.g.) TensorFlow versions at the same time?

Our goal is that the average user shouldn't need to worry about this. However, for users that do care, they can use the various `reticulate` tools for creating + using environments.

kevinushey:
Also worth saying that this machinery only activates when the user hasn't explicitly requested an environment. In that case, we automagically provide them with a Miniconda environment and use directives like these to prepare the environment as appropriate.

skeydan:
Oh, I see, that sound awesome! Thanks for the explanation!
So it seems it'll work great for new users as well as existing ones (who may be used to managing different environments) :-)

cboettig:
RE:
> unless you all think that there is so much breakage in the normal course of upgrades that we need to whitelist versions

Does your team test, say, the examples in the `keras` R package against each of the versions on GPU?  Last I tried (a few months ago now) we had a rough time getting many of the vignette examples you have in https://github.com/rstudio/keras/tree/master/vignettes/examples  to run on TF 2.0.0 (on cuda 10.0), but had better luck on older versions.  Though of course some of that could be user or hardware errors but it's made me nervous about compatibility issues in going to 2.0!  (likewise have seen a lot of breakage in random TF python scripts & libraries we've poked at when trying to move to TF 2.0).  

we're working on GPU-compatible docker images in [rocker](https://rocker-project.org), and it would be great to support things like the keras example vignettes out-of-the-box, so any guidance on the compatibility matrix you develop against there for TF version + cuda version would be most helpful! (and apologies, let me know if I should raise this in a separate thread instead? a dependency on TF 2.0.0 made me nervous!)

dfalbel:
@cboettig The examples were not yet tested against TensorFlow 2.0. We are in the process of updating the documentation to reflect all the 2.0 changes.

The R packages are tested against TF2.0  and we recommend using TF2.0 from now on. 

This PR doesn't add dependency on TF2. You can still use older versions of TensorFlow, but version 2.0 will be automatically installed if you are using the new `reticulate` automated installation machinery.

cboettig:
Thanks @dfalbel , this is great.  Do you test on GPU+CPU?  All cuda 10.0 now?

PR #377, opened 2019-11-22
Support for ignoring WARN and ERROR messages

javierluraschi:
Looks like the valid range is 0-3, not 0-1, see: https://github.com/tensorflow/tensorflow/blob/f15a1f88710ccda4a12bea9c945f4f44f4f102e3/tensorflow/core/platform/default/logging.h#L36-L39

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -42,7 +42,7 @@ tf_v2 <- function() {
   # but insist on printing warnings (level 2) and errors (level 3)
   cpp_log_opt <- getOption("tensorflow.core.cpp_min_log_level")
   if (!is.null(cpp_log_opt))
-    Sys.setenv(TF_CPP_MIN_LOG_LEVEL = max(min(cpp_log_opt, 1), 0))
+    Sys.setenv(TF_CPP_MIN_LOG_LEVEL = max(min(cpp_log_opt, 3), 0))
 
   # delay load tensorflow
   tf <<- import("tensorflow", delay_load = list(


skeydan:
This actually was on purpose... we didn't think it good to allow for disabling warnings and errors, be it only for support reasons - like, github issues that'd be tricky to work on if these kinds of messages are suppressed ...
Do you think there are cases where this is too restrictive on our side?

javierluraschi:
I would almost never suggest turning off `ERROR`, but turning off `WARNING` might be likely; sometimes systems get just wonky and might print a warnings that are ignorable... so say, if you were to have an R Markdown report, having a warning printing could be really annoying, same for info messages which is the case today.

Anyways, this is not a big deal cause I ended up changing the environment variable instead: https://github.com/mlverse/mlverse-docker/commit/4638d9f03e8b2097285802a044bd9a6c9055598d, so if anything, feels like maybe we should consider removing this entirely or my preference, defaulting to only `WARNING`.

When working in `sparklyr` we found out that we wanted to avoid being too verbose in the console; otherwise, users would get random print statements while you were typing in the console if something was happening in the background, which is really annoying, not sure if this happens with TF cpp code.

And yes, this is great for debugging but when the package is working correctly it can be really annoying. It's also easier to teach a few users how to troubleshoot than to teach everyone how to hide this message.

For instance, in the `mlverse` video from today: https://twitter.com/javierluraschi/status/1197892503099211778, there is a bunch of noise when I would much rather just have `tensorflow` render the Q&A answers sent to the GPT-2 model.

Feel free to close this one, or let me know if you want me to change the default behavior, we can also close and give this idea some more thought, the change is trivial so we can make this change any other chance.

One other random idea... I've been wanting to add a feature in RStudio... maybe a "log" pane or something like that... such that packages can redirect all diagnostics output there and users could easily copy-paste in support issues. I'll open a GitHub issue in RStudio as a start...

javierluraschi:
Actually, no need to implement this in the IDE, we have something similar in Spark in the connections UI. So we could build our own "TensorFlow Logs" feature by creating an RStudio add-in, maybe I'll volunteer myself to implement this.

PR #379, opened 2019-11-25
convert TensorShape to R lists

dfalbel:
Shapes a represented as instances of `TensorShape`:

``` r
library(tensorflow)
shape <- tf$TensorShape(list(NULL, 10L))
shape[0]
#> NULL
shape[1]
#> [1] 10
```

This PR will auto convert TensorShapes to lists so we can slice with 1-based indexing.

### DIFF
--- a/R/types.R
+++ b/R/types.R
@@ -13,3 +13,8 @@ shape <- function(...) {
       NULL
   })
 }
+
+#' @export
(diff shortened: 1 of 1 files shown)


skeydan:
LGTM - I was basically doing this manually in TFP tests...
I do think it can potentially break existing code that has already been adapted to TF2, but then it may be better to do it now than later?
Also, it would match expectations of those users who haven't yet been made aware ...

PR #395, opened 2020-01-19
tf 2.0 doc fix

alexanderbeatson:
Examples in the document are not updated. 
Starting from TF 2.0, we don't need 'session' anymore. 
I change the example codes with TF 2.0 compatibility.


dfalbel:
Thank you!

Issue #394, opened 2020-01-19
Make tfe_enable_eager_execution() Throws a message instead of error.


alexanderbeatson:
Please check R/eager.R
If eager execution is already running, it throws an error message 
> Error in py_get_attr_impl(x, name, silent) : 
>  AttributeError: module 'tensorflow' has no attribute 'contrib'

Can you add another if statement to determine eager execution is already running or not.
You can get True or False by running
> tf$executing_eagerly()
on TF 2.0 and later.

PR #397, opened 2020-02-07
Fixing contrib attribute not found error in tensorflow 2.0

alexanderbeatson:
As I mentioned the weird error message in [here](https://github.com/rstudio/tensorflow/issues/394), it throws error message from TensorFlow that it doesn't have the attribute name contrib. I added the stop function with a message that tells the user eager execution is already running. 

Please edit the error message at your convenience. 

Thank you.

### DIFF
--- a/R/eager.R
+++ b/R/eager.R
@@ -35,7 +35,10 @@
 #' @export
 tfe_enable_eager_execution <- function(
   config = NULL, device_policy = c("explicit", "warn", "silent")) {
-
+  # if eager execution is already running it will throw an error
+  if (tf$executing_eagerly()) {
+    stop('Tensorflow eager execution is already running')
+  }
   # alias eager mode (error if not available in this version of tf)
   contrib <- tf$contrib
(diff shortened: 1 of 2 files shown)


dfalbel:
IMHO we should just return `NULL` when eager execution is already enabled to follow the TensorFlow python API. 

I'm not sure in which version `tf$executing_eagerly()` was added to the main API we should also check to call the appropriate function.

alexanderbeatson:
@dfalbel 
As I checked other functions, they return error message using R stop() function. But, if you think returning NULL is better, may I request another pull request with returning NULL?

dfalbel:
@alexanderbeatson sure! IMHO users would not want to have an error if you are enabling eager execution an it's already running.

alexanderbeatson:
@dfalbel I just got an idea. Returning NULL alone would not make sense to some people. (They might think function is broken and they don't know eager is already running). So, I will return NULL with warning message. Thanks. Will make a pull request soon.

PR #398, opened 2020-02-11
[.tensor bug fix with all_dims()

t-kalinowski:
without this patch this results in an error:
```r
x <- tf$constant(1:10)
x[all_dims(), 3]
```


dfalbel:
Thanks very much @t-kalinowski!

PR #403, opened 2020-02-18
Add setuptools to apt packages to fix Travis build

terrytangyuan:
I observed the same error in my other project and this fixed the issue (hopefully fix the build here as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ addons:
       python3-pip
       python3-virtualenv
       python3-venv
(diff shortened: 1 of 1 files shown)


terrytangyuan:
@dfalbel Seems like this is not needed anymore since the CI has been migrated to use GitHub Workflows, right?

dfalbel:
@terrytangyuan Thanks very much! Yeah, we moved to Github Actions for the CI.

PR #408, opened 2020-02-25
tweaks to generics

### LINKED ISSUE #399: length() for scalar tensors should return 1
```
length(tf$constant(1L))
```
currently returns NULL, should probably return 1.

t-kalinowski:
- return correct `length()` for scalar tensors (closes #399 )
  - add `log1p()` method for tensors


dfalbel:
Thanks @t-kalinowski !

PR #468, opened 2021-06-08
update tests for generics

t-kalinowski:
- Refactor tests for generics
- add expm1
- add unary `+`

At some point, I remember I encountered an issue where the binary operators didn't dispatch correctly if the tensor was the second argument (that is, `tensor + 1` worked, but `1 + tensor` didn't). I couldn't reproduce this exact issue with the latest versions of everything, but I refactored the tests in case this pops up in the future 

While in there I also added support for expm1 and unary `+`.


tests/testthat/helper-utils.R
```diff
@@ -0,0 +1,46 @@
...
+
+.SESS <- NULL
+grab <- function(x) {
+  if(!inherits(x, "tensorflow.tensor"))
+    return(x)
+
+  if(tf$executing_eagerly())
+    return(as.array(x))
+
+  if (is.null(.SESS)) {
```
dfalbel, line 12: You are free to simplify this by dropping TF < 2.0 support.

dfalbel (approved):
Looks great! I didn't know of `helper-utils.R` when I first added this.

tests/testthat/helper-utils.R
t-kalinowski: Thanks! You're right. Will do on the next pass through.

PR #485, opened 2021-09-16
Tensor shapes

t-kalinowski:
- `shape()` now returns a `tf.TensorShape()` object
    (Previously an R-list of `NULL`s or integers).
- `[` method for `tf.TensorShape()` objects also now returns a `tf.TensorShape()`.
    Use `[[`, `as.numeric`, `as.integer`, and/or `as.list` to convert to R objects.
- `length()` method for `tensorflow.tensor` now returns `NA_integer_` for
    tensors with not fully defined shapes. (previously a zero length integer vector).
- `dim()` method for `tensorflow.tensor` now returns an R integer vector
    with `NA` for dimensions that are undefined.
    (previously an R list with `NULL` for undefined dimension)

- New S3 generics for `tf.TensorShape()`'s:
  `c`, `length`, `[<-`, `[[<-`, `merge`, `==`, `!=`,
  `as.list`, `as.integer`, `as.numeric`, `as.double`, `py_str`
  (joining previous generics `[` and `[[`).
  See `?shape` for extended examples.


dfalbel (approved):
LGTM!

PR #514, opened 2022-03-07
Expand Generics

t-kalinowski:


t-kalinowski:
Additions deferred for now:

- change `[` so it no longer passes tensors through as-is, but translates from 1-based to 0-based (This would help prevent nasty surprises when people wrap functions in `tf_function()`.
- expand `[` to accept integer vectors or integer arrays, which would dispatch to `tf.gather()` or `tf.gather_nd()` instead of `Tensor.__getitem__()`.
- `seq()` method wrapping `tf.range()`, but with R conventions of starting at 1 and including the requested end. This would pair with `[` treating tensors as 1 based. 

Additions considered but rejected:
- `as.list()` method wrapping `tf.unstack()`. It would need to flatten by default to match base behavior: `tf.unstack(tf.reshape(x, (-1,)))`. This would probably lead to more confusion than it's worth, so not adding.
- `xtfrm()` method wrapping `tf.argsort()`. Not sure how useful this would be in practice. What we really would want is `order()` or `rank()` to be generic, but those aren't generics and the docs suggest defining an `xtfrm()` method instead. However, with an `xtfrm()` method that returns a tensor, `order()` and `rank()` still don't work, so it's probably not worth it.

PR #519, opened 2022-03-13
make `[` 1-based for tensors

### LINKED ISSUE #516: `[` should translate Tensors from 1-based to 0-based

t-kalinowski:
closes #516

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # tensorflow (development version)
 
+- `[` method for tensors now interperts Tensors "R" style, 1 based and 
+  inclusive of the slice end. To restore the previous behavior, 
+  call `x[..., style = "python"]`, or set option `options(tensorflow.extract.style = "python").
+  
+- `[` gains the ability to accept a string for slice specs (optionally as a variable).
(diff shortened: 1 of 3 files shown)


t-kalinowski:
Still todo:
-  Update `[` tests so that all the `[` tests are evaluated in 3 contexts:
  with R values, with eager tensors, and with graph mode tensors in a `tf_function()`.
-  Update the warning and errors thrown when a `0` is encountered.

t-kalinowski:
Follow up: `tf.strided_slice()` does not accept tensors for `begin_mask`, `end_mask`, or `ellipsis_mask` arguments, only ints. This means that these arguments must be fully known at tracing time, and cannot be dependent on graph values. Consequently, building up a `tf.strided_slice()` call in the R `[.tensorflow.tensor` method is not a viable approach for us here.

Issue #605, opened 2024-10-08
Tensorboard fails to open if tensorflow prints warnings


RalphJO:
Hello,

In ```tensorflow:::tensorboard_version()```, if tensorflow gives any warnings the call will fail with ```Error: invalid version specification '<tensorflow warning text>'```

This comes from this line of the code:

https://github.com/rstudio/tensorflow/blob/b68aa4acf840c130c398590efdb47aea5cd9d1b3/R/tensorboard.R#L145

Please could you consider adding an ```ignore.stderr = TRUE``` to the system call so that warnings from tensorflow are not parsed, only the stdout from tensorboard?

```R
ver <- package_version(system("tensorboard --version_tb", intern = TRUE, ignore.stderr = TRUE)) 
```

Thank you,

Ralph

t-kalinowski:
Thanks for the bug report! Would you like to submit a PR?

PR #606, opened 2024-10-08
Issue #605 - ignore stderr when parsing tensorboard version

RalphJO:
As per issue #605, when getting the tensorboard version, only read the stdout from tensorboard, not the stderr.


t-kalinowski:
Thank you!

